PR #18, opened 2018-09-08
fix nil check

gcarling:


wcharczuk:
Closing b/c there is probably something upstream of this specific issue we need to fix.

PR #20, opened 2018-09-10
Allow healthz to prevent new connections before shutdown

mat285:


web/app.go
```diff
@@ -502,6 +523,17 @@ func (a *App) Start() (err error) {
 	keepAliveListener := TCPKeepAliveListener{a.listener}
 	var shutdownErr error
 	a.latch.Started()
+
+	// start the child app
+	if a.child != nil {
+		go func() {
+			err = a.child.Start()
```
wcharczuk: assigning `err` across goroutine boundaries is maybe not best practice, can you use a channel

mat285: yep will do

web/healthz.go
```diff
@@ -125,6 +130,20 @@ func (hz *Healthz) DefaultHeaders() map[string]string {
 	return hz.defaultHeaders
 }
 
+// Ready returns if healthz server is available ignoring the underlying server
+func (hz *Healthz) Ready() bool {
+	hz.readyLock.Lock()
```
wcharczuk: this is what the latches are for

mat285: 👍

wcharczuk:
It's a good start, but overall inverts the spirit of `healthz` which is as an add-on to an app to make it more production ready (where the app is just the stuff the user needs to build, and then slaps healthz on it)

wcharczuk:
sorry didn't mean to close this

PR #26, opened 2018-10-18
Follow the pointer during decomposition

mikes-blend:
Memory addresses aren't useful Strings.

### DIFF
--- a/util/reflection.go
+++ b/util/reflection.go
@@ -393,24 +393,11 @@ func (ru reflectionUtil) DecomposeStrings(obj interface{}, tagName ...string) ma
 				output[dataField] = typed
 			}
 		} else {
-			output[dataField] = fmt.Sprintf("%v", ru.followValuePointer(fieldValue))
+			output[dataField] = fmt.Sprintf("%v", ru.FollowValuePointer(fieldValue))
 		}
 	}
 
 	return output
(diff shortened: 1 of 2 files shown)


mikes-blend:
Need to handle nil better

mikes-blend:
Tested that nil behavior is now consistent with go-commons for better or worse.

util/reflection.go
```diff
@@ -393,13 +393,26 @@ func (ru reflectionUtil) DecomposeStrings(obj interface{}, tagName ...string) ma
...
-			output[dataField] = fmt.Sprintf("%v", fieldValue.Interface())
+			output[dataField] = fmt.Sprintf("%v", ru.followValuePointer(fieldValue))
 		}
 	}
 
 	return output
 }
 
+// followValuePointer derefs a reflectValue until it isn't a pointer, but will preseve it's nilness.
+func (ru reflectionUtil) followValuePointer(v reflect.Value) interface{} {
```
wcharczuk: did this function really not already exist?

amccarthy1: https://github.com/blend/go-sdk/blob/cf0e4b34712516647c7be247ee1c4a8725441ab6/util/reflection.go#L43-L54

I think it does...

PR #28, opened 2018-10-26
CreateMany from an empty slice just returns

mikes-blend:


db/invocation_test.go
```diff
@@ -237,6 +237,17 @@ func TestInvocationExistsError(t *testing.T) {
 	assert.NotNil(boolErr(conn.Invoke(context.Background()).WithLabel("exists_error_test").Exists(mustError)))
 }
 
+func TestInvocationCreateManyEmpty(t *testing.T) {
+	assert := assert.New(t)
+
+	var objs []uniqueObj = nil
```
wcharczuk: the `= nil` is unnecessary

PR #29, opened 2018-11-27
Fix and new method on Request

mikes-blend:
- Fixes an issue where the status code and other metadata are not returned from the request in the case of a 500.
- Adds a retrieval method for post data

### DIFF
--- a/request/request.go
+++ b/request/request.go
@@ -394,6 +394,13 @@ func (r *Request) WithPostData(field string, value string) *Request {
 	return r
 }
 
+// LookupPostData gets the first value set on the request via the WithPostData func associated with the given key.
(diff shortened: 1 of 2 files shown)


request/request.go
```diff
@@ -394,6 +394,13 @@ func (r *Request) WithPostData(field string, value string) *Request {
 	return r
 }
 
+// LookupPostData gets the first value set on the request via the WithPostData func associated with the given key.
+// If there are no values associated with the key, LookupPostData returns the empty string. This is a useful retrieval
+// mechanism for mocks
+func (r *Request) LookupPostData(field string) string {
```
wcharczuk: can you name this `GetPostData(...)`

mikes-blend:
addressed

PR #34, opened 2018-12-12
Make default values work in go-sdk/ask

mikes-blend:
Hit enter get default instead of hit enter get panic.

### DIFF
--- a/cmd/ask/main.go
+++ b/cmd/ask/main.go
@@ -108,9 +108,6 @@ func prompt(cv *configVar) {
 		prompt = fmt.Sprintf("%s: ", cv.Field)
 	}
 	cv.Value = sh.MustPrompt(prompt)
-	if len(cv.Value) == 0 && len(cv.Default) > 0 {
-		cv.Value = cv.Default
-	}
 }
 
 func secure(cv *configVar) {
--- a/cmd/ask/main.go
+++ b/cmd/ask/main.go
@@ -107,13 +107,9 @@ func prompt(cv *configVar) {
 	} else {
 		prompt = fmt.Sprintf("%s: ", cv.Field)
 	}
-	v, err := sh.Prompt(prompt)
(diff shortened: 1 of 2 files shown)


cmd/ask/main.go
```diff
@@ -108,6 +108,9 @@ func prompt(cv *configVar) {
 		prompt = fmt.Sprintf("%s: ", cv.Field)
 	}
 	cv.Value = sh.MustPrompt(prompt)
+	if len(cv.Value) == 0 && len(cv.Default) > 0 {
```
wcharczuk: again, this is superfluous, it's handled in `MarshalYAML`

sh/prompt.go
```diff
@@ -1,12 +1,19 @@
 package sh
 
-import "fmt"
+import (
+	"fmt"
+)
+
+const ErrUnexpectedNewline string = "unexpected newline"
```
wcharczuk: can you add a comment on this

sh/prompt.go
```diff
@@ -1,12 +1,19 @@
...
 
 // MustPrompt gives a prompt and reads input until newlines.
 // It panics on error.
 func MustPrompt(prompt string) string {
 	output, err := Prompt(prompt)
 	if err != nil {
+		if err.Error() == ErrUnexpectedNewline {
```
wcharczuk: you should be able to just do `exception.Is(...)` on this, there are a couple examples around

sh/prompt.go
```diff
@@ -17,5 +24,8 @@ func Prompt(prompt string) (string, error) {
 	fmt.Print(prompt)
 	var output string
 	_, err := fmt.Scanln(&output)
+	if err.Error() == ErrUnexpectedNewline {
```
wcharczuk: ditto comment on `exception.Is(...)` above

PR #35, opened 2019-01-04
Adding godoc badge

dhermes:
Also making the CircleCI badge link to the actual builds and dropping the invalid `circle-token=:circle-token` from the badge URL.


dhermes:
To see how it looks: https://github.com/dhermes/go-sdk/blob/patch-1/README.md

PR #37, opened 2019-01-10
Taggable interface for v1.0

amccarthy1:
Since this isn't a breaking change, would like to get it into the v1.0 branch as well


VERSION
```diff
@@ -1 +1 @@
-1.0.0
+1.0.1
```
amccarthy1: I did bump the version file

wcharczuk: cool

PR #39, opened 2019-01-25
Adds `jobkit`

wcharczuk:
This merge adds the `job` cli utility, as well as some jobkit helpers.

It also moves the `airbrake` and `datadog` packages to the root of the repository.


airbrake/notifier.go
```diff
@@ -0,0 +1,91 @@
...
+var (
+	defaultContextOnce sync.Once
+	defaultContext     map[string]interface{}
+)
+
+// New returns a new notifier.
+func New(cfg *Config) *Notifier {
+	// create a new reporter
+	client := gobrake.NewNotifierWithOptions(&gobrake.NotifierOptions{
+		ProjectId:   mustInt(cfg.ProjectID),
```
mat285: Not sure panicking here is best practice, `New` doesn't give any indication that it could crash a program misusing it

aws/ses/api_sender.go
```diff
@@ -0,0 +1,71 @@
...
+			Charset: &defaultCharset,
+			Data:    &m.TextBody,
+		}
+	}
+
+	_, err := s.client.SendEmailWithContext(ctx, input)
+	return exception.New(err)
+}
+
+func strPtrs(values []string) []*string {
```
mat285: As much as I love these functions, aws has these already `aws.StringSlice`

aws/config.go
```diff
@@ -0,0 +1,53 @@
...
+// Config is a config object.
+type Config struct {
+	Region          string `json:"region,omitempty" yaml:"region,omitempty" env:"AWS_REGION"`
+	AccessKeyID     string `json:"accessKeyID,omitempty" yaml:"accessKeyID,omitempty" env:"AWS_ACCESS_KEY_ID"`
+	SecretAccessKey string `json:"secretAccessKey,omitempty" yaml:"secretAccessKey,omitempty" env:"AWS_SECRET_ACCESS_KEY"`
+	SecurityToken   string `json:"securityToken,omitempty" yaml:"securityToken,omitempty" env:"AWS_SECURITY_TOKEN"`
+}
+
+// IsZero returns if the config is unset or not.
+func (a Config) IsZero() bool {
```
mat285: This is somewhat limiting on how to construct an aws client. We don't always need an accesskey/secretkey pair. The sdk supports using iam roles and ec2 permissions too

webutil/webhook.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"github.com/blend/go-sdk/exception"
+)
+
+// Webhook is a configurable request.
+type Webhook struct {
+	Method  string            `json:"method" yaml:"method"`
+	URL     string            `json:"url" yaml:"url"`
+	Headers map[string]string `json:"headers" yaml:"headers"`
+	Body    string            `json:"body" yaml:"body"`
```
mat285: It doesn't look like this is used in the `Send` function?

aws/config.go
wcharczuk: Do we want to save those values to config files though? The `ses` package specifically has been changed to take a session as the constructor argument (not a config) so it should be more generally usable. 

The spirit of this config file is to help with loading key pairs, we should probably have setup logic to prefer config, and then default to the aws session setup code.

sh/tee.go
```diff
@@ -0,0 +1,28 @@
+package sh
+
+import "io"
+
+// Tee retruns a new tee writer for a given set of writers.
```
mat285: `returns`

stringutil/spit_space_quoted.go
```diff
@@ -51,3 +50,17 @@ func SplitSpaceQuoted(text string) (output []string) {
 	}
 	return
 }
+
+func isQuote(r rune) bool {
+	return r == '"' || r == '\'' || r == '“' || r == '”' || r == '`'
```
mat285: haha yeah ran into these dumb slack quotes too

jobkit/management_server_test.go
```diff
@@ -0,0 +1 @@
+package jobkit
```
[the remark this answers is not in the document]
wcharczuk: resolved

sh/tee.go
wcharczuk: 👀

webutil/webhook.go
wcharczuk: It does not, added a test for this.

PR #41, opened 2019-02-19
Logger: Increase padded with from 27 to 30 for timestamps.

dhermes:
H/T to https://stackoverflow.com/a/47165010/1068170

### DIFF
--- a/logger/text_writer_test.go
+++ b/logger/text_writer_test.go
@@ -86,7 +86,7 @@ func TestFormatTimestamp(t *testing.T) {
 	pstTz, err := time.LoadLocation("America/Los_Angeles")
 	assert.Nil(err)
 
-	tsValues := [8]time.Time{
+	tsValues := [9]time.Time{
 		time.Date(2019, time.February, 19, 15, 12, 47, 123000000, time.UTC),
 		time.Date(2019, time.February, 19, 15, 12, 47, 123400000, time.UTC),
 		time.Date(2019, time.February, 19, 15, 12, 47, 123450000, time.UTC),
@@ -94,9 +94,10 @@ func TestFormatTimestamp(t *testing.T) {
 		time.Date(2019, time.February, 19, 15, 12, 47, 123456700, time.UTC),
 		time.Date(2019, time.February, 19, 15, 12, 47, 123456780, time.UTC),
 		time.Date(2019, time.February, 19, 15, 12, 47, 123456789, time.UTC),
+		time.Date(2019, time.February, 19, 15, 12, 47, 1234567890, time.UTC),
 		time.Date(2019, time.February, 19, 7, 12, 47, 123456789, pstTz),
 	}
-	expectedLogs := [8]string{
+	expectedLogs := [9]string{
 		"2019-02-19T15:12:47.123Z       [error] test string\n",
 		"2019-02-19T15:12:47.1234Z      [error] test string\n",
 		"2019-02-19T15:12:47.12345Z     [error] test string\n",
@@ -104,6 +105,7 @@ func TestFormatTimestamp(t *testing.T) {
 		"2019-02-19T15:12:47.1234567Z   [error] test string\n",
 		"2019-02-19T15:12:47.12345678Z  [error] test string\n",
 		"2019-02-19T15:12:47.123456789Z [error] test string\n",
+		"2019-02-19T15:12:48.23456789Z  [error] test string\n", // Note seconds has increased
 		"2019-02-19T07:12:47.123456789-08:00 [error] test string\n",
 	}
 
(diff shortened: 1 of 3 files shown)


dhermes:
@wcharczuk Feel free to just say "no I don't want this". I sent it because when the text logger is used for "realtime human consumption" we occasionally get logs that don't line up, which is a bit jarring, e.g.

```
2019-02-19T06:08:10.838609912Z [task-runner] [info] Kubernetes magic
2019-02-19T06:08:10.83862185Z [task-runner] [info] Also magic
2019-02-19T06:08:10.853861217Z [task-runner] [info] Also k8s
```

wcharczuk:
Hey @dhermes, it's an interesting approach, what we do ultimately though is just right pad the string when we write out. Can see it here: https://github.com/blend/go-sdk/blob/master/logger/text_writer.go#L220, ideally 27 characters is enough to cover the timestamp, but this may need some adjusting

dhermes:
@wcharczuk The 27 character width is for millisecond precision, which is the default on many operation systems. See https://play.golang.org/p/tm0TacSCr7u for some different timestamps that make the length go beyond 27. ISTM we'd want to optimize for full nanos with UTC, which needs 30 chars?

logger/text_writer_test.go
```diff
@@ -79,3 +80,36 @@ func TestWriterLabels(t *testing.T) {
...
+	assert := assert.New(t)
+
+	tsValues := [7]time.Time{
+		time.Date(2019, time.February, 19, 15, 12, 47, 123000000, time.UTC),
+		time.Date(2019, time.February, 19, 15, 12, 47, 123400000, time.UTC),
+		time.Date(2019, time.February, 19, 15, 12, 47, 123450000, time.UTC),
+		time.Date(2019, time.February, 19, 15, 12, 47, 123456000, time.UTC),
+		time.Date(2019, time.February, 19, 15, 12, 47, 123456700, time.UTC),
+		time.Date(2019, time.February, 19, 15, 12, 47, 123456780, time.UTC),
+		time.Date(2019, time.February, 19, 15, 12, 47, 123456789, time.UTC),
```
wcharczuk: i would do `int(math.MaxInt64)` for one of these nano values if you really want to push some bounds

logger/text_writer_test.go
```diff
@@ -79,3 +80,40 @@ func TestWriterLabels(t *testing.T) {
...
+	}
+
+	for i, ts := range tsValues {
+		expected := expectedLogs[i]
+		buffer := bytes.NewBuffer(nil)
+		writer := NewTextWriter(buffer)
+		writer.showTimestamp = true
+		writer.useColor = false
+		writer.WriteError(Messagef(Error, "test %s", "string").WithTimestamp(ts))
+		assert.Equal(string(buffer.Bytes()), expected)
```
wcharczuk: buffer has a `.String()` method

logger/text_writer_test.go
dhermes: But that isn't a valid timestamp?

logger/text_writer_test.go
```diff
@@ -79,3 +80,40 @@ func TestWriterLabels(t *testing.T) {
...
+	}
+
+	for i, ts := range tsValues {
+		expected := expectedLogs[i]
+		buffer := bytes.NewBuffer(nil)
+		writer := NewTextWriter(buffer)
+		writer.showTimestamp = true
+		writer.useColor = false
+		writer.WriteError(Messagef(Error, "test %s", "string").WithTimestamp(ts))
+		assert.Equal(buffer.String(), expected)
```
dhermes: @wcharczuk Using `buffer.String()` here. Do you want me to also modify other uses of `string(buffer.Bytes())` in this file?

wcharczuk: nah i can do that in another pass

logger/text_writer_test.go
wcharczuk: https://play.golang.org/p/zl8WY9cPhQ4

it's interesting, it actually causes the timestamp to move forward, which makes sense as times are just `struct{int64,int64}` and adding a bunch to the nanos would just add more to the seconds fracttion int64 since epoch

wcharczuk: the point of that was to make sure to have as many decimal places in the rfc3339nanos formatted ts as possible, b/c that's ultimately what we're sniffing for with this test, that we don't give the ts enough padding in the text output

dhermes: Yup. I have that, [it's 9 digits][1]: `"2006-01-02T15:04:05.999999999Z07:00"`

[1]: https://github.com/golang/go/blob/go1.11.4/src/time/format.go#L79

dhermes: OK, I added a 10 digit integer which will "overflow" the digit count

PR #42, opened 2019-02-23
Fix final coverage being written to root 

gcarling:


wcharczuk:
instead of just removing the file write, i would add a flag that disables root coverage calculation + file write, but also we need to fix the total calculation

PR #43, opened 2019-02-27
Set up AutotriggerAction

alexfish8:
- Thoughts on naming this `AutotriggerAction` vs `AutotriggerHandler`?


async/auto_action.go
```diff
@@ -0,0 +1,130 @@
...
+
+import (
+	"sync"
+	"time"
+
+	"github.com/blend/go-sdk/exception"
+)
+
+// NewAutoAction returns a new NewAutoAction
+func NewAutoAction(interval time.Duration) *AutoAction {
```
wcharczuk: wouldn't you also need the max count as a required input?

async/auto_action.go
```diff
@@ -0,0 +1,130 @@
...
+				a.Trigger()
+			}
+			a.latch.Stopped()
+			return
+		}
+	}
+}
+
+// Set sets the value
+func (a *AutoAction) SetValue(value interface{}) {
```
wcharczuk: i would refactor this to be `Increment` and have it `atomic.AddInt32(&a.counter, 1)`

wcharczuk: basically it avoids grabbing the lock to increment, but would want to grab the lock to force the trigger. you can use `atomic.Load` to do the test if after the increment we'd need to trigger (before doing the normal add) there are some patterns in the stdlib and online for how to do increments like this

async/auto_action.go
```diff
@@ -0,0 +1,162 @@
...
+func (a *AutoAction) TriggerAsync() {
+	a.valueLock.Lock()
+	defer a.valueLock.Unlock()
+	a.triggerUnsafeAsync()
+}
+
+// triggerUnsafe calls the handler, if one is set, without acquiring any locks
+func (a *AutoAction) triggerUnsafe() {
+	if a.handler != nil {
+		a.handler(a.value)
```
wcharczuk: you can really just have the `handler` or `listener` be a closure and have it implicit it manages a value.

```
var foo string
auto := new(AutoAction)
auto.Handler = func() {
    ext.Send(&foo)
})
foo = "bar"
auto.Handler() // sends "bar"
```
etc.

async/auto_action.go
```diff
@@ -9,12 +9,11 @@ import (
...
 
 // NewAutoAction returns a new NewAutoAction
-func NewAutoAction(interval time.Duration, maxCount int32) *AutoAction {
+func NewAutoAction(interval time.Duration, maxCounter int32) *AutoAction {
 	return &AutoAction{
-		value:          nil,
-		valueLock:      &sync.Mutex{},
-		maxCount:       maxCount,
-		handler:        nil,
+		Mutex:          sync.Mutex{},
```
wcharczuk: you don't need to do this specifically; if you embed the mutex it's implicit

alexfish8: ah, right

PR #44, opened 2019-03-04
Added slack methods to send messages and read the response.

nellamad:


slack/response.go
```diff
@@ -0,0 +1,21 @@
+package slack
+
+// Response is a slack response
+type Response struct {
+	OK              bool            `json:"ok"`
+	Channel         string          `json:"channel,omitempty"`
+	ThreadTimestamp string          `json:"ts,omitempty"`
+	Message         ResponseMessage `json:"message,omitempty"`
```
nellamad: Considered reusing `Message` instead of `ResponseMessage` here but thought it would be cleaner to keep them separate since they differ in certain fields such as `thread_ts` in `Message` and `ts` in `ResponseMessage`.

slack/response.go
```diff
@@ -0,0 +1,21 @@
+package slack
+
+// Response is a slack response
+type Response struct {
```
wcharczuk: might want to make this a more specific name, maybe `PostMessageResponse` or similar

slack/response.go
wcharczuk: I would keep them the same so there are fewer types to update / manage. If a type has a couple potentially duplicate looking fields, might be good to add comments on the fields to that effect (i.e. what they're used for)

slack/webhook_sender.go
```diff
@@ -61,7 +63,28 @@ func (whs WebhookSender) Send(ctx context.Context, message Message) error {
 	return nil
 }
 
-// PostMessage posts a basic message to a given chanel.
+// SendAndRead sends a slack hook and returns the deserialized response
+func (whs WebhookSender) SendAndRead(ctx context.Context, message Message) (*Response, error) {
```
wcharczuk: I would add `Response` to the end, i.e. `SendAndReadResponse`, would also do this for `PostMessageAndRead(Response)`

slack/message.go
```diff
@@ -27,6 +27,12 @@ type Message struct {
 	ReplyBroadcast  bool                `json:"reply_broadcast"`
 	LinkNames       int                 `json:"link_names"`
 	Attachments     []MessageAttachment `json:"attachments"`
+
+	// Response-specific fields
+	BotID     string `json:"bot_id,omitempty"`
+	Type      string `json:"type,omitempty"`
+	SubType   string `json:"bot_message,omitempty"`
```
wcharczuk: SubType vs. `bot_message` is this intended?

nellamad: Not intended, good thing we have extra eyeballs here.

nellamad: For context/clarity, I pulled these response fields from [this doc page](https://api.slack.com/methods/chat.postMessage). Not sure if there is anything more comprehensive that would list more fields to add.

wcharczuk: as long as we have the fields we need to do threaded responses we can add more over time (afaik the slack response types are kind of fluid and add fields regularly)

PR #46, opened 2019-03-07
Increase millisecond metric precision.

alexfish8:


stats/mock_collector.go
```diff
@@ -76,5 +76,5 @@ type MockMetric struct {
 }
 
 func millis(d time.Duration) float64 {
```
wcharczuk: we should use the `timeutil.Millis` function if possible for this

alexfish8:
The `go-sdk/logger` package has a Milliseconds function. Opting to create a duplicate function so as to not create a dependency between `go-sdk/logger` and `go-sdk/timeutil`, but can refactor accordingly if we want to prioritize code de-duplication over dependency flattening.

PR #49, opened 2019-03-27
Mikes blend/add list to secrets

mikes-blend:
Add a "List" operation to the secrets interface


request/constants.go
```diff
@@ -34,6 +34,8 @@ const (
 	MethodDelete = "DELETE"
 	// MethodOptions is a method.
 	MethodOptions = "OPTIONS"
+	// MethodList is a method.
+	MethodList = "LIST"
```
mikes-blend: Can take this out if you want. These aren't the constants that secrets uses

secrets/mock_client.go
```diff
@@ -45,3 +46,28 @@ func (c *MockClient) Delete(_ context.Context, key string, options ...Option) er
 	delete(c.SecretValues, key)
 	return nil
 }
+
+// List lists keys on a path
+func (c *MockClient) List(_ context.Context, path string, options ...Option) ([]string, error) {
```
mikes-blend: Was harder to mock than implement 🙄

request/constants.go
wcharczuk: AFAIK it's not a standard HTTP verb, I would omit.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods

secrets/mock_client.go
wcharczuk: Yikes no kidding

request/constants.go
mikes-blend: I figured it was in the extended registry, but it is not. Can’t believe hashicorp just made their own shit up.

http://www.iana.org/assignments/http-methods/http-methods.xhtml

PR #51, opened 2019-04-05
Use db.Config.Schema to build the psql search_path

mikes-blend:
Concerns with this: Its technically a breaking change if people are expecting Schema to not impact the search_path and don't have their DB User to have the specified schema front their search path.

### DIFF
--- a/db/config.go
+++ b/db/config.go
@@ -39,6 +39,8 @@ func NewConfigFromDSN(dsn string) (*Config, error) {
 			config.Password = strings.TrimPrefix(piece, "password=")
 		} else if strings.HasPrefix(piece, "sslmode=") {
 			config.SSLMode = strings.TrimPrefix(piece, "sslmode=")
+		} else if strings.HasPrefix(piece, "search_path=") {
+			config.Schema = strings.TrimPrefix(piece, "search_path=")
 		}
 	}
 	return &config, nil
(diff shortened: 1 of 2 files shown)


db/config.go
```diff
@@ -241,9 +243,16 @@ func (c Config) CreateDSN() string {
 		return c.GetDSN()
 	}
 
-	var sslMode string
+	var queryString string
+	var qParam = "?"
 	if len(c.GetSSLMode()) > 0 {
-		sslMode = fmt.Sprintf("?sslmode=%s", url.QueryEscape(c.GetSSLMode()))
+		queryString = fmt.Sprintf("%s%ssslmode=%s", queryString, qParam, url.QueryEscape(c.GetSSLMode()))
```
wcharczuk: i've since refactored this method to use `url.URL` to build the DSN, can you please rework this to use that basis.

wcharczuk:
https://www.postgresql.org/docs/10/libpq-connect.html

I'm also not sure you can set the search path on the DSN. I used to do this as an immediate first step after opening the connection as an explicit `Exec("set search_path=$1", c.Config.SearchPath)`. 

I'm curious if this works (setting it on the DSN).

mikes-blend:
It does work. The ability to configure run time parameters in the query string is one of the improvements of golang's pq over lib/pq on which it is based.

>Similarly to libpq, when establishing a connection using pq you are expected to supply a connection string containing zero or more parameters. A subset of the connection parameters supported by libpq are also supported by pq. Additionally, pq also lets you specify run-time parameters (such as search_path or work_mem) directly in the connection string. This is different from libpq, which does not allow run-time parameters in the connection string, instead requiring you to supply them in the options parameter.

mikes-blend:
> can you pull the latest changes and add to the config tests

I'll rebase, but I've already added testing for the search path to the config tests. What additional tests would you like to see there?

wcharczuk:
nvm, this is against 2.0

PR #53, opened 2019-04-11
Add Transit Encryption to Vault Client

akarimcheese:
- Adds Transit Interface
- Makes Client interface a union of Transit and KV interfaces
- Implement Vault Transit


secrets/client.go
```diff
@@ -4,8 +4,13 @@ import "context"
 
 // Client is the general interface for a Secrets client
 type Client interface {
+	// KV
 	Put(ctx context.Context, key string, data Values, options ...RequestOption) error
```
wcharczuk: can you separate these out into `KVClient` and `TransitClient`.

```
type KVClient interface {
    Put(ctx context.Context, key string, data Values, options ...RequestOption) error
    Get(ctx context.Context, key string, options ...RequestOption) (Values, error)
    Delete(ctx context.Context, key string, options ...RequestOption) error
    List(ctx context.Context, path string, options ...RequestOption) ([]string, error)
}
```

and 
```
type TransitClient interface {
    Encrypt(ctx context.Context, key string, context, data []byte) (string, error)
    Decrypt(ctx context.Context, key string, context []byte, ciphertext string) ([]byte, error)
}
```

And have `Client` be:

```
 type Client interface {
    KVClient
    TransitClient
}
```

secrets/client.go
```diff
@@ -4,8 +4,13 @@ import "context"
...
 	Get(ctx context.Context, key string, options ...RequestOption) (Values, error)
 	Delete(ctx context.Context, key string, options ...RequestOption) error
 	List(ctx context.Context, path string, options ...RequestOption) ([]string, error)
+
+	// Transit
+	TransitEncrypt(key string, context map[string]interface{}, data []byte) (string, error)
```
wcharczuk: can you make sure to include `ctx context.Context` as the first arg for these too.

secrets/client.go
wcharczuk: note that you can embed interfaces. i think you already have the separate transit client interface, so it would just be a matter of pulling those methods out of the parent interface and using interface embedding.

secrets/vault_transit.go
```diff
@@ -0,0 +1,84 @@
...
+	_ Transit = VaultTransit{}
+)
+
+// VaultTransit defines vault transit interactions
+type VaultTransit struct {
+	Client *VaultClient
+}
+
+// TransitEncrypt encrypts a given set of data.
+func (vt VaultTransit) TransitEncrypt(key string, context map[string]interface{}, data []byte) (string, error) {
```
wcharczuk: context should just be bytes, we can add helpers above this to create a context from a json bag or similar.

akarimcheese:
would you prefer i rename KV -> KVClient, or is the alias I put in (KVClient = KV) fine?

PR #54, opened 2019-04-12
Rework Guards to respect schemas

mikes-blend:
Guards that work with schemas.


db/config.go
```diff
@@ -193,7 +193,7 @@ func (c Config) GetDatabase(inherited ...string) string {
 
 // GetSchema returns the connection schema or a default.
 func (c Config) GetSchema(inherited ...string) string {
-	return configutil.CoalesceString(c.Schema, "", inherited...)
+	return configutil.CoalesceString(c.Schema, DefaultSchema, inherited...)
```
wcharczuk: did you mean `c.DefaultSchema()` here?

db/connection.go
```diff
@@ -134,6 +135,15 @@ func (dbc *Connection) StatementInterceptor() StatementInterceptor {
 	return dbc.statementInterceptor
 }
 
+// DefaultSchema returns the schema that this connection has at the front of its search_path. If none is specified in
+// the config, this will be 'public'
+func (dbc *Connection) DefaultSchema() string {
+	if dbc.config == nil || strings.TrimSpace(dbc.config.Schema) == "" {
```
wcharczuk: it should be sufficient to say `dbc.config.schema == ""`

db/migration/guard_predicates.go
```diff
@@ -1,51 +1,107 @@
...
-	return TableNotExistsWithPredicate(PredicateTableExists, tableName)
+	return GuardNotPredicate(fmt.Sprintf("Check table does not exist: %s", tableName), PredicateTableExists, tableName)
 }
 
-// IndexNotExists creates a index on the given connection if it does not exist.
-func IndexNotExists(tableName, indexName string) GuardFunc {
-	return IndexNotExistsWithPredicate(PredicateIndexExists, tableName, indexName)
+// TableExistsInSchema returns a guard that ensures a table exists
+func TableExistsInSchema(schemaName, tableName string) GuardFunc {
+	return GuardPredicate2(fmt.Sprintf("Check table exists: %s.%s", schemaName, tableName),
```
wcharczuk: instead of having `GuardPredicate2` i would have (1) predicate type `func() (bool, error)` and return curried versions of that with `predicate` and `predicate2` etc.

wcharczuk: so `PredicateColumnExists, ..., ...` would be come `PredicateColumnExists(table, column)` itself returning a func() (bool, error)

db/config.go
mikes-blend: No. I want this to default explicitly to "public" if not set. DefaultSchema() is a helper on the connection. This is setting a default for the config.

db/migration/guard_predicates.go
mikes-blend: What is the advantage you see in doing this? I feel like the PredicateN and GuardPredicateN stuff should just be unexported and resigned to "package internals" anyways. External users should just use Guard as the "Predicate" stuff doesn't seem that useful unless you're writing a whole suite of GuardFunc implementations like we are here.

I don't see a whole lot of gain to currying these either, because it's not really making things any more abstract/simple/generic. Its' the same number of methods afaict, just now with more function wrapping.

mikes-blend:
@wcharczuk So this PR has kind of turned into a dumping ground for all the fixes I've needed for profile. Where do you want me to go with this? Merge it? Break it up? Make it on master and do a backport to v2?

wcharczuk:
honestly i would fix glaring issues / bugs w/ v2.0, but for profile service i would work off master. no hard promises, but not many breaking changes (from me anyway) coming for master

mikes-blend:
I'm almost certainly going to port to v3 when we cut it. I'll just close the PR and keep a profile branch in the interim.

PR #55, opened 2019-04-17
Encrypt then mac on stream encryptor

mat285:
The current approach uses Encrypt and Mac instead of Encrypt then Mac. While its a small difference the later is the only one provably secure so we should use it instead. https://en.wikipedia.org/wiki/Authenticated_encryption#Approaches_to_authenticated_encryption
https://crypto.stackexchange.com/questions/202/should-we-mac-then-encrypt-or-encrypt-then-mac


crypto/stream_test.go
```diff
@@ -12,7 +12,7 @@ func TestStreamEncryptorDecryptor(t *testing.T) {
 	assert := assert.New(t)
 	key, err := CreateKey(32)
 	assert.Nil(err)
-	plaintext := "Eleven is the best person in all of Hawkins Indiana. KSHVdyveduytvadsguvdsjgcv"
```
mat285: I must have written this like right after stranger things came out 😂

PR #61, opened 2019-05-03
Adds tests to get `r2`

wcharczuk:
This adds a significant amount of tests to the `r2` package.

It also slightly refactors the `OptDial` options to be more collaborative, otherwise they would clobber each other if you set multiple.


r2/opt_dial.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"net"
+	"net/http"
+)
+
+// OptDial sets dial options for a request, these must be done all at once.
+func OptDial(opts ...DialOption) Option {
+	return func(r *Request) error {
+		if r.Client == nil {
```
gcarling: the code setting up r.Client and r.Client.Transport seems to be used in all these options, could maybe it factor it out somewhere. seems like a lot of the options share a kind of base constructor, curious if this is a thing in other packages

gcarling: could you put this stuff here? https://github.com/blend/go-sdk/pull/61/files#diff-bffb747d535990946f7212343f6edd0aR18 and then it'll be initialized when it reaches the options

gcarling: that link didn't go through - it was supposed to be to the `New` func that takes the options, but i guess since there's no diff on that line the link didn't function as intende

wcharczuk: yeah a lot of this is repetitive, the goal is to keep the raw (Request) object usable w/o pointer receivers

PR #64, opened 2019-05-09
Stats Top Level Tests

akarimcheese:
## PR Summary

Unit tests for top level files in go-sdk/stats

 - **Type:** Internal
 - **Issue Link:** 

#### Reviewers:


stats/mock_collector.go
```diff
@@ -28,7 +28,7 @@ type MockCollector struct {
 }
 
 // AddDefaultTag adds a default tag.
-func (mc MockCollector) AddDefaultTag(key, value string) {
+func (mc *MockCollector) AddDefaultTag(key, value string) {
```
mat285: good catch

PR #65, opened 2019-05-09
Tests for the Web Package

mat285:
## PR Summary

Adding tests to the web package. Bringing coverage over 80%

 - **Type:** Internal

#### Reviewers:
@gcarling @mikes-blend 


web/config_test.go
```diff
@@ -1 +1,182 @@
...
+	assert := assert.New(t)
+	c := &Config{}
+	assert.Equal(DefaultSessionTimeout, c.SessionTimeoutOrDefault())
+	c.SessionTimeout = 10
+	assert.Equal(c.SessionTimeout, c.SessionTimeoutOrDefault())
+}
+
+func TestConfigCookieNameOrDefault(t *testing.T) {
+	assert := assert.New(t)
+	c := &Config{}
```
wcharczuk: nit; `var c Config`

https://dave.cheney.net/2014/05/24/on-declaring-variables not sure if this is a pattern we want to follow but might be counter intuitive

web/config_test.go
```diff
@@ -1 +1,182 @@
...
+	yes := true
+	assert.Equal(DefaultCookieSecure, c.CookieSecureOrDefault())
+	c.BaseURL = "https://hello.com"
+	assert.True(c.CookieSecureOrDefault())
+	c.BaseURL = "http://hello.com"
+	assert.False(c.CookieSecureOrDefault())
+	c.CookieSecure = &yes
+	assert.Equal(*c.CookieSecure, c.CookieSecureOrDefault())
```
gcarling: `assert.True`? seems odd to change format here. i guess its consistent with the rest of the tests in this file

web/view_cache_config_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/blend/go-sdk/env"
+
+	"github.com/blend/go-sdk/assert"
+)
+
+func TestViewCacheConfigResolve(t *testing.T) {
+	assert := assert.New(t)
+	vcc := &ViewCacheConfig{}
+
+	env.SetEnv(env.New())
```
gcarling: you can `defer env.Restore()` here

web/config_test.go
mat285: So I wrote it this way because I don't care whether they are true or false here, just that they are the same. Which is what the functionality should be and makes it easier if tests change in the future

web/config_test.go
```diff
@@ -1 +1,182 @@
 package web
+
+import (
+	"encoding/base64"
+	"testing"
+
+	"github.com/blend/go-sdk/env"
```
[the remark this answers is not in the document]
mat285: done

PR #74, opened 2019-05-24
Schema Awareness and Singleton Removal for db and migrations

mikes-blend:
## PR Summary

This PR adds awareness around Schemas to the db and db/migration packages. The end results of this patch are that setting a schema in your DB config now automatically sets your search_path to include that schema, precluding you from needing to do custom configuration of your ROLE or your connection once it is set up.

This PR also adds the ability to migrate schemas other than the one set in the search path and to easily create guards for these migrations. It also changes the guard implementation queries for several guards to ones that are a bit more standard (and in the case of constraint is actually correct). This has to be a major (backwards-breaking) change, as the old signature for ConstraintExists did not take enough parameters to uniquely identify a constraint and would have performed incorrectly had two tables in a DB had a constraint with the same name.

Finally, this PR removes the Default() singleton used by the db package. Services should use their own method of managing their connections. This is also a backwards compatibility breaking change. 

 - **Type:** Features
 - **Issue Link:**

#### Reviewers:


db/default.go
```diff
@@ -1,30 +1,23 @@
 package db
 
-import (
-	"sync"
-)
-
 var (
 	defaultConnection *Connection
-	defaultLock       = sync.Mutex{}
```
mikes-blend: This mutex is Radioactive Man’s googles:
https://i.kym-cdn.com/photos/images/facebook/000/021/347/1248075250110.jpg

db/config.go
```diff
@@ -33,8 +33,8 @@ func NewConfigFromDSN(dsn string) (*Config, error) {
 			config.Password = strings.TrimPrefix(piece, "password=")
 		} else if strings.HasPrefix(piece, "sslmode=") {
 			config.SSLMode = strings.TrimPrefix(piece, "sslmode=")
-		} else if strings.HasPrefix(piece, "schema=") {
```
mikes-blend: Schema is not the real arg we wanted here. search_path, when provided to go/lib-pq will actually set the search path on the connection. Neither the schema arg nor the search_path arg do a damn thing in a psql DSN, but pq behaves differently:
> Similarly to libpq, when establishing a connection using pq you are expected to supply a connection string containing zero or more parameters. A subset of the connection parameters supported by libpq are also supported by pq. Additionally, pq also lets you specify run-time parameters (such as search_path or work_mem) directly in the connection string. This is different from libpq, which does not allow run-time parameters in the connection string, instead requiring you to supply them in the options parameter.

https://godoc.org/github.com/lib/pq

db/context.go
```diff
@@ -0,0 +1,18 @@
+package db
+
+import "context"
+
+type skipQueryLogging struct {}
+
+// WithSkipQueryLogging sets the context to skip logger listener triggers.
+func WithSkipQueryLogging(ctx context.Context) context.Context {
```
mikes-blend: Should this be moved to an invocation Option? Or is it ok to do this with a context key?

Makefile
```diff
@@ -1,5 +1,7 @@
 PREFIX			?= $(shell pwd)
 PKGS 			:= $(shell go list ./... | grep -v /vendor/)
+# We don't lint yaml because its forked code and terrible at lint
+LINTPKGS        := $(shell go list ./... | grep -v /vendor/ | grep -v "go-sdk/yaml")
```
wcharczuk: 👏

db/config.go
```diff
@@ -144,6 +144,14 @@ func (c Config) DatabaseOrDefault(inherited ...string) string {
 	return DefaultDatabase
 }
 
+// SchemaOrDefault returns the schema on the search_path or the default ("public")
+func (c Config) SchemaOrDefault(inherited ...string) string {
```
wcharczuk: Should we change this to `SearchPathOrDefault` and any other references to schema to be more consistent?

db/context.go
wcharczuk: logger has a similar thing (a context based option to kill logging) we should probably just use that.

wcharczuk:
is this really a major version change?

db/config.go
mikes-blend: I'm torn on this because it moves us even more towards supporting postgres exclusively. We could rename everything including the Config field to SearchPath, which in the current state-of-affairs is probably most accurate. IMHO I've found multi-schema search paths in postgres to be more trouble than they are worth. Interschema naming collisions tend to occur and are difficult to track down. I feel that we should support a single default schema on the search path and only have the library support single-schema paths, unless you really want to munge it otherwise and understand what you're doing. If we take that approach leaving the "Schema" name makes more sense. I can be convinced either way here.

mikes-blend:
> is this really a major version change?

Technically any PR that changes the signature of an exported method should be considered major. This PR does 3 major change things:

- Adds a required parameter of table_name to the ConstraintExists and ConstraintNotExists GuardFuncs, so that constraints with the same name on different tables (which postgres allows) don't confuse the guards. This breaks the backwards compatibility of these guards
- Unexports the predicate builders. Though I doubt anyone was actually using them externally to the package anyways, that is a backwards compatibility breaking change
- Removes Default(), OpenDefault(), and SetDefault()

I have assumed that until we "cut" v3.0 that while major changes deserve more scrutiny to ensure we are changing things in the "right way," they are permitted. The goal of 3.0 seems to be to ship the "best" thing we can, regardless of compatibility constraints.

db/context.go
mikes-blend: I need this to kill both the metrics and the logging, since the trace logs some query info that might contain sensitive parameters in the case where the statement cannot be parameterized.

db/config.go
wcharczuk: woof, got it, so if we ever switched the underlying driver this would stop working

db/config.go
wcharczuk: I'm OK with this library `go-sdk/db` being basically exclusively for postgres. It's baked into a bunch of stuff already (CreateMany, Upsert, all of the migration sniffs).

wcharczuk: We should bring that larger point to the group though.

db/config.go
mikes-blend: Yeah. I'd say if we did change the driver we'd implement this in some other way

PR #75, opened 2019-05-26
Adds more logger tests. Tweaks some options with the logger.

wcharczuk:
## PR Summary

 - **Type:** Internal
 - **Issue Link:** N/A

#### Reviewers:


logger/COVERAGE
```diff
@@ -1 +1 @@
-46.9
\ No newline at end of file
+39.81
```
wcharczuk: @gcarling wat

mikes-blend: 🤔

logger/default.go
```diff
@@ -6,48 +6,48 @@ import (
 )
 
 var (
-	log     *Logger
-	logInit sync.Once
+	_log     *Logger
```
mikes-blend: What is the convention for prepending underscores to var names?

wcharczuk:
> When did the logger singleton come back? I thought we were trying to move away form the singleton pattern within the SDK libs?

it got added back as a compatibility shim for paul (it mirrors the api for sirupsen/logrus). we can remove

logger/default.go
wcharczuk: in this case, it avoids a conflict with the stdlib `log` package. the linter (golint) allows it, but we probably shouldn't promulgate it as a pattern.

PR #76, opened 2019-05-28
Migrations: options and name overhaul

mikes-blend:
## PR Summary

Adds Options for migration Suites and Groups. Renames types to values that match their NewBlah functions. Increases test coverage.

 - **Type:** Improvements
 - **Issue Link:** 

#### Reviewers:


db/migration/action.go
```diff
@@ -34,17 +34,17 @@ func Statements(statements ...string) Action {
 
 // Exec runs a statement with a given set of arguments.
```
wcharczuk: I would add another line here
> It is intended to be used as a migration action in lieu of the typical `migration.Statements(...)`

db/migration/group.go
```diff
@@ -8,37 +8,53 @@ import (
 	"github.com/blend/go-sdk/ex"
 )
 
-// Group creates a new GroupedActions from a given list of actionable.
-func Group(actions ...Actionable) GroupedActions {
-	return GroupedActions(actions)
+// NewGroup creates a new GroupedActions from a given list of actionable.
+func NewGroup(options ...GroupOption) *GroupedActions {
+	g := &GroupedActions{}
```
wcharczuk: so, i don't know why, but sometimes this will not work, and it's better to do

```
var g GroupedActions
for _, opt := range options {
    opt(&g)
}
return &g
```

db/migration/group.go
```diff
@@ -8,37 +8,53 @@ import (
...
+	for _, o := range options {
+		o(g)
+	}
+	return g
+}
+
+// NewWithActions is a helper for "migrations.NewGroup(migrations.OptActions(actions ...migration.Actionable))"
+func NewWithActions(actions ...Actionable) *GroupedActions {
```
wcharczuk: `NewGroupWithActions`?

db/migration/group.go
```diff
@@ -8,37 +8,53 @@ import (
...
+	return NewGroup(OptActions(actions...))
 }
 
 // GroupedActions is an atomic series of migration actions.
 // It uses transactions to apply these actions as an atomic unit.
-type GroupedActions []Actionable
+type GroupedActions struct{
```
wcharczuk: `Group`?

db/migration/group.go
```diff
@@ -8,37 +8,53 @@ import (
...
 
 // Action runs the groups actions within a transaction.
-func (ga GroupedActions) Action(ctx context.Context, c *db.Connection) (err error) {
-	var tx *sql.Tx
-	tx, err = c.Begin()
-	if err != nil {
-		return
+func (ga *GroupedActions) Action(ctx context.Context, c *db.Connection) (err error) {
+	var tx *sql.Tx = nil
+	if !ga.NoTransaction {
```
wcharczuk: So, this is correct in that the default intended value is `false` i.e. use a transaction, i would name it `SkipTransaction` or similar

db/migration/group.go
mikes-blend: 🤔 Will do. Seems strange

db/migration/group.go
mikes-blend: Yep. Typo on my part

db/migration/group.go
mikes-blend: Will do

db/migration/suite.go
```diff
@@ -10,21 +10,28 @@ import (
 )
 
 // New returns a new suite of groups.
-func New(groups ...GroupedActions) *Suite {
-	return &Suite{
-		Groups: groups,
+func New(options ...SuiteOption) *Suite {
+	s := Suite{}
```
wcharczuk: var s Suite

db/migration/suite.go
```diff
@@ -10,21 +10,28 @@ import (
...
 type Suite struct {
 	Log    logger.Log
-	Groups []GroupedActions
+	Groups []*Group
 
-	Applied int
-	Skipped int
-	Failed  int
-	Total   int
+	applied int
```
wcharczuk: might as well make these exported

db/migration/suite.go
```diff
@@ -10,21 +10,28 @@ import (
...
+// NewWithGroups is a helper for "New(OptGroups(groups ...*Group))"
+func NewWithGroups(actions ...*Group) *Suite {
+	return New(OptGroups(actions...))
 }
 
 // Suite is a migration suite.
 type Suite struct {
 	Log    logger.Log
-	Groups []GroupedActions
+	Groups []*Group
```
wcharczuk: why do these need to be pointer receivers?

db/migration/suite.go
mikes-blend: I just didn't want someone external to be able to modify them. Shouldn't "operational" fields like this be unexported to protect against inadvertent writes.

db/migration/suite.go
mikes-blend: Done to maintain the existing convention of NewBlah returns a *Blah. So NewGroup returns a *Group which is collected here. GroupedActions used to be a collection type so this also isn't making something that wasn't nillable nillable.

db/migration/action.go
```diff
@@ -9,21 +9,20 @@ import (
...
+func NoOp(_ context.Context, _ *db.Connection, _ *sql.Tx) error { return nil }
 
 // Statements returns a body func that executes the statments serially.
 func Statements(statements ...string) Action {
-	return func(ctx context.Context, c *db.Connection, tx *sql.Tx, options ...db.InvocationOption) (err error) {
-		opts := append([]db.InvocationOption{db.OptContext(ctx), db.OptTx(tx)}, options...)
+	return func(ctx context.Context, c *db.Connection, tx *sql.Tx) (err error) {
 		for _, statement := range statements {
-			err = c.Invoke(opts...).Exec(statement)
+			err = c.Invoke(db.OptContext(ctx), db.OptTx(tx)).Exec(statement)
```
wcharczuk: we could maybe create a helper, `ActionOpts(ctx, tx) []db.InvocationOption`

db/migration/suite.go
wcharczuk: I've taken as a default stance with everything else to export fields because you never know when something else might need to do this and it's much better to export by default in those cases.

db/migration/action.go
mikes-blend: Its clunky because of the slice vs. variadics thing. I feel like this is tolerable as is.

PR #77, opened 2019-05-29
Adds a way to shortcut db invocations with an error.

wcharczuk:
## PR Summary

Allows db invocations to be pre-empted by an error on construction. Not useful in the sdk now, but useful in libraries that wrap it.

 - **Type:** Improvements
 - **Issue Link:** N/A

#### Reviewers:

### DIFF
--- a/db/invocation.go
+++ b/db/invocation.go
@@ -23,6 +23,7 @@ type Invocation struct {
 	TraceFinisher        TraceFinisher
 	StartTime            time.Time
 	Tx                   *sql.Tx
+	Err                  error
 }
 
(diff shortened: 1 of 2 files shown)


db/invocation.go
```diff
@@ -852,6 +853,9 @@ func (i *Invocation) CloseStatement(stmt *sql.Stmt, err error) error {
 
 // Start runs on start steps.
 func (i *Invocation) Start(statement string) (string, error) {
+	if i.Err != nil {
```
mikes-blend: Do we want an OptErr?

wcharczuk: for now it's not a usecase we really need to support so my sense is we don't need it

PR #80, opened 2019-06-05
DB Invocation method return refactor

mikes-blend:
## PR Summary

This started out trying to solve for the fact that Query.Out() is leaving junk in structs it writes to, if the new values from the db are "zero values."

It turned into a refactor of a bunch of behaviors in the db package around what is returned from things. The highlights:

For method return signatures:
`Out()`, `Delete()`, `Update()`, and `Get()` are changing to (bool, err) where the bool indicates an object has been found/deleted/updated. They also error if more than one matching row is found. `ErrTooManyRows`
`Exec()` is changing to (int, err) where the int is the number of rows affected if applicable
`Create()`, `CreateMany()`,`Upsert()` and, `CreateIfNotExists()` remain the same
`Truncate()` helper has been removed

`Out()` and `Get()` now scrub the structs they populate when no rows are returned. They also scrub any db-tagged fields in the struct that are not in the result set, or are in the result set but have "empty" values.

`Into()` is a new method on Query that can be used to chain populate a struct. It does not clear any of the aforementioned data.  

 - **Type:** Bugfix and Refactor
 - **Issue Link:** 

#### Reviewers:

### DIFF
--- a/db/column.go
+++ b/db/column.go
@@ -72,11 +72,7 @@ func (c Column) SetValue(object interface{}, value interface{}) error {
 
 	valueReflected := ReflectValue(value)
 	if !valueReflected.IsValid() {
-		if field.Kind() == reflect.Ptr {
-			field.Set(reflect.Zero(field.Type()))
-		} else {
-			field.Set(reflect.Zero(ReflectType(value)))
-		}
+		field.Set(reflect.Zero(field.Type()))
 		return nil
 	}
 
--- a/db/column.go
+++ b/db/column.go
@@ -72,7 +72,11 @@ func (c Column) SetValue(object interface{}, value interface{}) error {
 
 	valueReflected := ReflectValue(value)
 	if !valueReflected.IsValid() {
-		field.Set(reflect.Zero(ReflectType(value)))
+		if field.Kind() == reflect.Ptr {
+			field.Set(reflect.Zero(field.Type()))
+		} else {
+			field.Set(reflect.Zero(ReflectType(value)))
+		}
 		return nil
 	}
 
--- a/db/column.go
+++ b/db/column.go
@@ -61,6 +61,14 @@ type Column struct {
 	Inline       bool
 }
 
+// EmptyValue sets the empty value on a field on a database mapped object
+func (c Column) EmptyValue(object interface{}) error {
+	objValue := ReflectValue(object)
+	field := objValue.FieldByName(c.FieldName)
+	field.Set(reflect.Zero(field.Type()))
+	return nil
+}
+
 // SetValue sets the field on a database mapped object to the instance of `value`.
 func (c Column) SetValue(object interface{}, value interface{}) error {
 	objValue := ReflectValue(object)
--- a/db/column.go
+++ b/db/column.go
@@ -61,8 +61,8 @@ type Column struct {
 	Inline       bool
 }
 
-// EmptyValue sets the empty value on a field on a database mapped object
-func (c Column) EmptyValue(object interface{}) {
+// SetZero sets the empty value on a field on a database mapped object
+func (c Column) SetZero(object interface{}) {
 	objValue := ReflectValue(object)
 	field := objValue.FieldByName(c.FieldName)
 	field.Set(reflect.Zero(field.Type()))
--- a/db/column.go
+++ b/db/column.go
@@ -72,6 +72,7 @@ func (c Column) SetValue(object interface{}, value interface{}) error {
 
 	valueReflected := ReflectValue(value)
 	if !valueReflected.IsValid() {
+		field.Set(reflect.Zero(ReflectType(value)))
 		return nil
 	}
 
--- a/db/column.go
+++ b/db/column.go
@@ -62,15 +62,14 @@ type Column struct {
 }
 
 // EmptyValue sets the empty value on a field on a database mapped object
-func (c Column) EmptyValue(object interface{}) error {
+func (c Column) EmptyValue(object interface{}) {
 	objValue := ReflectValue(object)
 	field := objValue.FieldByName(c.FieldName)
 	field.Set(reflect.Zero(field.Type()))
-	return nil
 }
 
 // SetValue sets the field on a database mapped object to the instance of `value`.
-func (c Column) SetValue(object interface{}, value interface{}) error {
+func (c Column) SetValue(object interface{}, value interface{}, clearEmpty bool) error {
 	objValue := ReflectValue(object)
 	field := objValue.FieldByName(c.FieldName)
 	fieldType := field.Type()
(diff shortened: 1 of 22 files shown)


db/column.go
```diff
@@ -72,6 +72,7 @@ func (c Column) SetValue(object interface{}, value interface{}) error {
 
 	valueReflected := ReflectValue(value)
 	if !valueReflected.IsValid() {
+		field.Set(reflect.Zero(ReflectType(value)))
```
mikes-blend: My reflection-foo is poor. Is this the right way to achieve this?

mikes-blend:
Working on a test that replicates and tests the issue.

db/column.go
mikes-blend: This is wrong because pointers. Working on it

db/column.go
```diff
@@ -61,8 +61,15 @@ type Column struct {
 	Inline       bool
 }
 
+// EmptyValue sets the empty value on a field on a database mapped object
+func (c Column) EmptyValue(object interface{}) {
```
wcharczuk: Can we make this `SetZero` to be consistent with `SetValue`

db/column.go
```diff
@@ -61,8 +61,15 @@ type Column struct {
...
+	objValue := ReflectValue(object)
+	field := objValue.FieldByName(c.FieldName)
+	field.Set(reflect.Zero(field.Type()))
+}
+
 // SetValue sets the field on a database mapped object to the instance of `value`.
-func (c Column) SetValue(object interface{}, value interface{}) error {
+func (c Column) SetValue(object interface{}, value interface{}, clearEmpty bool) error {
```
wcharczuk: why have both `EmptyValue` and `Set(...,clearEmpty:true)`?

db/connection.go
```diff
@@ -210,12 +210,12 @@ func (dbc *Connection) DefaultSchema() string {
 }
 
 // Exec is a helper stub for .Invoke(...).Exec(...).
-func (dbc *Connection) Exec(statement string, args ...interface{}) error {
+func (dbc *Connection) Exec(statement string, args ...interface{}) (int, error) {
```
wcharczuk: How does this work for RDBMS that support multiple statements per call?

db/invocation.go
```diff
@@ -78,96 +83,29 @@ func (i *Invocation) Query(statement string, args ...interface{}) *Query {
 }
 
 // Get returns a given object based on a group of primary key ids within a transaction.
-func (i *Invocation) Get(object DatabaseMapped, ids ...interface{}) (err error) {
+func (i *Invocation) Get(object DatabaseMapped, ids ...interface{}) (found bool, err error) {
```
wcharczuk: one hesitation; is there a way to design this api to preserve the existing behavior and add this bool incrementally? even if we have to change the name of the existing function, that's an easier replace than adding a new return value

db/invocation.go
```diff
@@ -78,96 +83,29 @@ func (i *Invocation) Query(statement string, args ...interface{}) *Query {
...
-		if err != nil {
-			err = Error(err)
-			return
-		}
-
-		objValue := ReflectValue(obj)
-		collectionValue.Set(reflect.Append(collectionValue, objValue))
-	}
-	return
+	return i.Query(queryBody).OutMany(collection)
```
wcharczuk: one subtlety that isn't captured here. `Get` and `All` populated _in order_ and out many populates _by name_. This doesn't have a huge impact but it does affect performance. If you know what the column order is ahead of time you skip many map reads to populate each row.

db/column.go
mikes-blend: Yes

db/column.go
mikes-blend: EmptyValue is a helper method to just empty the tagged field of a struct. The clearEmpty flag is used in SetValue to determine whether or not to call EmptyValue

db/connection.go
mikes-blend: Will investigate. I am under the impression that it will return the number of statements executed in the last call. does lib/pq support multistatement calls?

wcharczuk: it's adding support for it soon; crdb supports multiple statements, afaik psql is adding support soon

db/invocation.go
mikes-blend: Do you want two versions like Get and GetWithMetadata? I can do that pretty trivially.

wcharczuk: eh leave it as is, we can add a helper later

db/invocation.go
mikes-blend: I did notice this. I understand the efficiency loss, but I think that I'd rather take the latency for the sake of a reduction in code complexity. Can we calculate the latency of the map lookups in a 10 column 1000 row result set? Is it still on the order of microseconds there, or does it tick up into millis. If it does I'll add the inOrder back in

db/connection.go
mikes-blend: The driver, not our code determines the number. I don't think we can do "better than lib/pq" in this case.

db/invocation.go
mikes-blend: I added a helper like this for Delete in my manager.

db/invocation.go
wcharczuk: based on this only ever being used by all, lets leave the populator in case people want to use it, but keep the code cleanup stuff

db/populate.go
```diff
@@ -46,34 +57,6 @@ func PopulateByName(object interface{}, row Rows, cols *ColumnCollection) error
 	return nil
 }
 
-// PopulateInOrder sets the values of an object in order from a sql.Rows object.
-// Only use this method if you're certain of the column order. It is faster than populateByName.
-// Optionally if your object implements Populatable this process will be skipped completely, which is even faster.
-func PopulateInOrder(object DatabaseMapped, row Scanner, cols *ColumnCollection) (err error) {
```
wcharczuk: Please add this back

db/invocation.go
```diff
@@ -54,10 +54,15 @@ func (i *Invocation) Exec(statement string, args ...interface{}) (err error) {
...
 
-	if _, err = stmt.ExecContext(i.Context, args...); err != nil {
+	res, err := stmt.ExecContext(i.Context, args...)
+	if err != nil {
 		err = Error(err)
 		return
 	}
+	// The error here is intentionally ignored. Postgres supports this. We'd need to revisit swallowing this error
+	// for other drivers
+	ra64, _ := res.RowsAffected()
```
wcharczuk: (1) we should handle / capture that error and float it to the top if it's set. don't just swallow it
(2) we should change the return type to mirror rows affected, i.e i64

mikes-blend: 1) I am worried that operations that don't support row counts (e.g. CREATE ROLE) may have an error in here
2) Fine.... but who uses 32 bit systems anymore anyways.

wcharczuk: yeah agreed int aliased to i64 makes this moot but better to be consistent

PR #81, opened 2019-06-05
Creates `GZip` middleware for `go-sdk/web`

wcharczuk:
## PR Summary

Moves GZip compression from being always on, gated by a header sniff, to be something you have to specifically add to the app as middleware.

 - **Type:** Improvements|Internal
 - **Issue Link:** n/a

#### Reviewers:


webutil/gzip_response_writer.go
```diff
@@ -0,0 +1,77 @@
...
+type GZipResponseWriter struct {
+	gzipWriter    *gzip.Writer
+	innerResponse http.ResponseWriter
+	statusCode    int
+	contentLength int
+}
+
+// InnerResponse returns the underlying response.
+func (crw *GZipResponseWriter) InnerResponse() http.ResponseWriter {
+	return crw.innerResponse
```
gcarling: why not just have this field be public if we are returning it here?

webutil/gzip_response_writer.go
```diff
@@ -0,0 +1,77 @@
...
+// Write writes the byes to the stream.
+func (crw *GZipResponseWriter) Write(b []byte) (int, error) {
+	_, err := crw.gzipWriter.Write(b)
+	crw.contentLength += len(b)
+	return len(b), err
+}
+
+// Header returns the headers for the response.
+func (crw *GZipResponseWriter) Header() http.Header {
+	return crw.innerResponse.Header()
```
gcarling: why have these accessor methods if the field is returned in `InnterResponse` - just common use cases?

PROFANITY_RULES.yml
```diff
@@ -3,6 +3,11 @@ GOLANG_LOGGER: #note: this does not affect go-sdk/logger because of the package
   contains:
   - "*logger.Logger"
 
+GOLANG_LATCH:
+  description: "please use the latch constructor"
+  contains:
+  - "&async.Latch{}"
```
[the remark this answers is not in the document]
wcharczuk: yeah fair

webutil/gzip_response_writer.go
wcharczuk: so it can satisfy an interface; there are no interfaces for fields unfortunately

webutil/gzip_response_writer.go
wcharczuk: interfaces

webutil/gzip_response_writer.go
mat285: Should we have a way to set the inner response since we can't mess with it directly? (same with header)

mat285: Actually its probably fine without it

PR #82, opened 2019-06-07
Adding `Forbidden()` and `Unauthorized()` result providers.

dhermes:
## PR Summary

 - **Type:** Features

#### Reviewers:

- I was surprised to find out that `web.JSONResultProvider.NotAuthorized()` actually returns a 403 Forbidden

### DIFF
--- a/web/json_result_provider.go
+++ b/web/json_result_provider.go
@@ -30,6 +30,14 @@ func (jrp JSONResultProvider) NotAuthorized() Result {
 	}
 }
 
+// Forbidden returns a 403 Forbidden response.
+func (jrp JSONResultProvider) Forbidden() Result {
+	return &JSONResult{
+		StatusCode: http.StatusForbidden,
+		Response:   "Forbidden",
+	}
+}
+
 // InternalError returns a service response.
 func (jrp JSONResultProvider) InternalError(err error) Result {
 	if err != nil {
--- a/web/json_result_provider.go
+++ b/web/json_result_provider.go
@@ -25,7 +25,7 @@ func (jrp JSONResultProvider) NotFound() Result {
 // NotAuthorized returns a service response.
 func (jrp JSONResultProvider) NotAuthorized() Result {
 	return &JSONResult{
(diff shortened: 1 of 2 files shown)


dhermes:
Seems like CircleCI choked

dhermes:
> imho we could trim this down to just changing the status code returned by `Unauthorized`

Did you mean the already existing `NotAuthorized()`? If yes, I agree with you (but didn't want to introduce a breaking change).

dhermes:
@wcharczuk Thoughts?

web/json_result_provider.go
```diff
@@ -30,6 +30,22 @@ func (jrp JSONResultProvider) NotAuthorized() Result {
 	}
 }
 
+// Forbidden returns a 403 Forbidden response.
+func (jrp JSONResultProvider) Forbidden() Result {
```
mikes-blend: I don't think we should have both Forbidden and NotAuthorized. I think we should just make this "Forbidden" to match http norms and not backport it to earlier releases to avoid breaking backwards compatibility there. There are already a lot of backwards-compat breaking changes pending for V3. Let's just pull this in with them.

@dhermes Do yo need Unauthorized back-ported to v2?

dhermes: @mikes-blend Thanks for the reply. I sent this more out of "`NotAuthorized` is wrong" than out of anything else. I'm perfectly fine just calling

```go
	&web.JSONResult{
		StatusCode: http.Status...,
		Response:   "...",
	}
```

in my code that uses `go-sdk/web`.

dhermes:
@wcharczuk @mikes-blend Mind CR-ing?

mikes-blend:
Can you please move this from a fork to a branch, so that it unborks the automated checks? After that I'll review.

dhermes:
Closed this in favor of #84. At some point @wcharczuk (or someone else with CircleCI admin) should allow PR builds from forks: https://circleci.com/gh/blend/go-sdk/edit

PR #83, opened 2019-06-16
Adds `go-sdk/cache`

wcharczuk:
## PR Summary

Adds a `go-sdk/cache` package. This package contains an interface for key/value cache implementations to conform to, and an implementation of a simple local memory cache.

Implementation details:

- The local cache is a normal map with a `sync.RWMutex` synchronizing access to it.
- The `Sweep` method makes use of a priority queue, ordered by the expiration times of the key/values stored in the map as pointers.
- The `Sweep` process is, it grabs an exclusive lock, peeks an element and removes it if the expiry is in the past. If it reaches an element with an expiry in the future, it stops.
- The local cache allows values to be `Fix`'d, or have their expiry times updated, but this is very expensive in large cardinality. 

As a result of the above, the local cache is heavily optimized for adding new elements quickly, getting elements quickly, and expiring elements on a ttl. 

It is *NOT* optimized for updating the expiry of elements (though it allows it) nor is it optimized for removing elements on demand.

It includes an example as well.

 - **Type:** Features

#### Reviewers:


cache/lru_queue.go
```diff
@@ -0,0 +1,261 @@
...
+
+// NewLRUQueue creates a new, empty, LRUQueue.
+func NewLRUQueue() *LRUQueue {
+	return &LRUQueue{
+		array: make([]*Value, ringBufferDefaultCapacity),
+	}
+}
+
+// LRUQueue is a fifo buffer that is backed by a pre-allocated array, instead of allocating
+// a whole new node object for each element (which saves GC churn).
```
mat285: How big of a problem is GC churn relative to the added cost of O(n) over O(1)

wcharczuk: it's pretty significant, there are benchmarks in the collections package

PR #85, opened 2019-06-19
Secrets Refactor

mikes-blend:
## PR Summary

Refactor and Fixes in go-sdk/secrets:

Timeout parameter on vault client is now respected.

KV clients now correctly return map[string]interface{} not map[string]string, matching 
capabilities with vault KV2

APM Tracing client for secrets

Option pattern for Transit Key Create/Update

 - **Type:** Improvements and Bugfix

#### Reviewers:


mikes-blend:
This is still in progress: Looking to collect early feedback on timeout handling, tracing, and some secrets API changes.

secrets/mock_client.go
```diff
@@ -14,7 +14,8 @@ var (
 
 // MockTransitKey is a mocked transit key
 type MockTransitKey struct {
-	Properties map[string]interface{}
+	CreationProps *TransitKeyCreation
```
wcharczuk: 1) these probably don't need to be pointers
2) i'm not sure what these do from their names

wcharczuk: if they're pointers to test if the've been set or not, can probably work around that

secrets/mock_client_test.go
```diff
@@ -93,7 +93,7 @@ func TestMockClientTransitEncrypt(t *testing.T) {
 	assert := assert.New(t)
 	client := NewMockClient()
 
-	err := client.CreateTransitKey(context.TODO(), "key1", map[string]interface{}{"mock_option": true})
+	err := client.CreateTransitKey(context.TODO(), "key1", TransitKeyCreation{})
```
wcharczuk: this is an uncommon pattern in the repo; what are the motivations for using options singletons vs. the options pattern?

mikes-blend: I was thinking about that this weekend. I'll take a stab at using options here.

secrets/mock_client.go
```diff
@@ -84,38 +85,35 @@ func (c *MockClient) List(_ context.Context, path string, options ...RequestOpti
 }
 
 // CreateTransitKey creates a new transit key.
-func (c *MockClient) CreateTransitKey(ctx context.Context, key string, params map[string]interface{}) error {
+func (c *MockClient) CreateTransitKey(ctx context.Context, key string, params TransitKeyCreation) error {
 	c.TransitKeys[key] = MockTransitKey{
-		Properties: make(map[string]interface{}),
 		Keys:       make(map[string][]byte),
```
akarimcheese: did you intend to assign params?

secrets/mock_client.go
```diff
@@ -84,38 +85,35 @@ func (c *MockClient) List(_ context.Context, path string, options ...RequestOpti
...
-		keyPath.Properties[opt] = value
-	}
+	keyPath.UpdateProps = &config
 
 	c.TransitKeys[key] = keyPath
 	return nil
 }
 
 // ReadTransitKey returns data about a transit key path
 func (c *MockClient) ReadTransitKey(ctx context.Context, key string) (map[string]interface{}, error) {
```
akarimcheese: should we define a type for what reading the transit key will return?

secrets/mock_client.go
```diff
@@ -84,38 +85,35 @@ func (c *MockClient) List(_ context.Context, path string, options ...RequestOpti
...
-	keyPath, ok := c.TransitKeys[key]
+	_, ok := c.TransitKeys[key]
 	if !ok {
 		return map[string]interface{}{}, fmt.Errorf("No key")
 	}
 
-	return keyPath.Properties, nil
+	return map[string]interface{} {"SampleIntProp": 1, "SampleBoolProp": true, "SampleStringProp": "Foo"}, nil
```
akarimcheese: would it be reasonable to return properties set from creation/update?

secrets/tracer.go
```diff
@@ -0,0 +1,15 @@
+package secrets
+
+import (
+	"net/http"
+)
+
+// Tracer is a tracer for requests.
+type Tracer interface {
+	Start(*http.Request) TraceFinisher
```
akarimcheese: is it reasonable to extend this to allow for implementations to tag/name traces by:
- key name
- encrypt/decrypt?
- get/put (KV) create/update/read/delete (transit keys)?

stats/tracing/secretstrace/tracer.go
```diff
@@ -0,0 +1,47 @@
...
+// Tracer returns a request tracer that also injects span context into outgoing headers.
+func Tracer(tracer opentracing.Tracer) secrets.Tracer {
+	return &secretsTracer{tracer: tracer}
+}
+
+type secretsTracer struct {
+	tracer opentracing.Tracer
+}
+
+func (st secretsTracer) Start(req *http.Request) secrets.TraceFinisher {
```
akarimcheese: would be nice to add a resource name to the trace span

secrets/tracer.go
mikes-blend: It tags it with the vault API url path being hit, which contains most of this info

wcharczuk:
> Replace this text with a description of what this PR is accomplishing

can you add more color here

secrets/transit_opts.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// TKCreateOption is an option type for transit key creation
+type TKCreateOption func(tkc *TKCreateConfig) error
+
+// TKUpdateOption is an option type for transit key creation
+type TKUpdateOption func(tkc *TKUpdateConfig) error
+
+// TKCreateOptConfig is a creation option for when you have a pre-defined struct
+func TKCreateOptConfig(config TKCreateConfig) TKCreateOption {
```
wcharczuk: can we keep the `Opt...` format, it helps with drilling into intellisense / autocorrect lists.

secrets/transit_opts.go
```diff
@@ -0,0 +1,138 @@
...
+	return func(tkc *TKCreateConfig) error {
```
wcharczuk: `CreateTransitKeyOptions`

secrets/transit_opts.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+}
+
+// TKUpdateOptMinDecryptionVersion - Specifies the minimum version of ciphertext allowed to be decrypted. Adjusting this
+// as part of a key rotation policy can prevent old copies of ciphertext from being decrypted, should they fall into
+// the wrong hands. For signatures, this value controls the minimum version of signature that can be verified
+// against. For HMACs, this controls the minimum version of a key allowed to be used as the key for verification.
+func TKUpdateOptMinDecryptionVersion(minDecryptionVersion int) TKUpdateOption {
+	return func(tku *TKUpdateConfig) error {
+		tku.MinDecryptionVersion = minDecryptionVersion
```
wcharczuk: are there sanity checks for this? can the min be < 0? we should error here if possible

secrets/vault_transit.go
```diff
@@ -18,18 +18,18 @@ type VaultTransit struct {
 }
 
 // CreateTransitKey creates a transit key path
-func (vt VaultTransit) CreateTransitKey(ctx context.Context, key string, params map[string]interface{}) error {
-	req := vt.Client.createRequest(MethodPost, filepath.Join("/v1/transit/keys/", key)).WithContext(ctx)
-
-	if _, ok := params["type"]; !ok {
-		params["type"] = "aes256-gcm96"
+func (vt VaultTransit) CreateTransitKey(ctx context.Context, key string, options ...TKCreateOption) error {
+	config := &TKCreateConfig{}
```
wcharczuk: style nit;
`var config CreateTransitKeyOptions`
and then use `&config` saves having things like the json body take a pointer

secrets/vault_transit.go
```diff
@@ -45,10 +45,18 @@ func (vt VaultTransit) CreateTransitKey(ctx context.Context, key string, params
...
+		err := o(config)
+		if err != nil {
+			return err
+		}
+	}
+
 	req := vt.Client.createRequest(MethodPost, filepath.Join("/v1/transit/keys/", key, "config")).WithContext(ctx)
 
-	body, err := vt.Client.jsonBody(config)
+	body, err := vt.Client.jsonBody(*config)
```
wcharczuk: also avoids stuff like this

stats/tracing/secretstrace/tracer.go
```diff
@@ -0,0 +1,62 @@
...
```
wcharczuk: One thing to verify with this; how do other implementations format these options? is vault a separate application? is it just an external call for the existing application? etc.

secrets/transit_opts.go
mikes-blend: There are a large number of scenarios that can cause this to error at vault. I figured it would be better to send the request and let the vault API reject it than try to catch all the failure conditions here.

stats/tracing/secretstrace/tracer.go
mikes-blend: I need to do some digging on this.

secrets/constants.go
```diff
@@ -43,3 +43,13 @@ const (
 	// Version2 is a constant.
 	Version2 = "2"
 )
+
+// These types are encryption algorithms that can be used when creating a transit key
+const (
```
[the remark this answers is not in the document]
mikes-blend: I had that to start with, but it just made everything harder for no real gain. I think the godocs are explicit enough. If you strongly object I can put the type back.

secrets/tracer.go
mikes-blend: I'll look at refactoring this a bit

secrets/transit.go
```diff
@@ -6,8 +6,8 @@ import (
 
 // TransitClient is an interface for an encryption-as-a-service client
 type TransitClient interface {
```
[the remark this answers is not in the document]
mikes-blend: I think that we should either get rid of the secrets.client interface altogether or leave it the way it is. The former is a larger refactor and I didn't think it was worth it, as it makes migrating existing implementations that much harder.

stats/tracing/secretstrace/tracer.go
mikes-blend: I've not really been able to find good examples of "other implementations" Do you have suggestions?

secrets/kv1.go
```diff
@@ -24,7 +24,7 @@ func (kv1 KV1) Put(ctx context.Context, key string, data Values, options ...Requ
 	}
 	req := kv1.Client.createRequest(MethodPut, filepath.Join("/v1/", key), options...).WithContext(ctx)
 	req.Body = contents
-	res, err := kv1.Client.send(req)
+	res, err := kv1.Client.send(req, TraceOptVaultOperation("kv1.put"), TraceOptKeyName(key))
```
wcharczuk: `OptTrace`

secrets/transit_opts.go
```diff
@@ -0,0 +1,138 @@
...
+	return func(tku *UpdateTransitKeyConfig) error {
+		tku.MinDecryptionVersion = minDecryptionVersion
+		return nil
+	}
+}
+
+// TKUpdateOptMinEncryptionnVersion - Specifies the minimum version of the key that can be used to encrypt plaintext,
+// sign payloads, or generate HMACs. Must be 0 (which will use the latest version) or a value greater or equal to
+// min_decryption_version.
+func TKUpdateOptMinEncryptionnVersion(minEncryptionVersion int) UpdateTransitKeyOption {
```
akarimcheese: dup?

mikes-blend: Oops

secrets/vault_transit.go
```diff
@@ -18,24 +18,24 @@ type VaultTransit struct {
 }
 
 // CreateTransitKey creates a transit key path
-func (vt VaultTransit) CreateTransitKey(ctx context.Context, key string, params map[string]interface{}) error {
-	req := vt.Client.createRequest(MethodPost, filepath.Join("/v1/transit/keys/", key)).WithContext(ctx)
-
-	if _, ok := params["type"]; !ok {
```
akarimcheese: are we recording breaking changes anywhere? we should include this (and the derived default value) in any such list. not aware of any that exists though

stats/tracing/secretstrace/tracer.go
```diff
@@ -0,0 +1,72 @@
...
+		return
+	}
+	if vaultError != nil {
+		tracing.SpanError(stf.span, vaultError)
+	}
+	stf.span.SetTag(tracing.TagKeyHTTPCode, strconv.Itoa(vaultStatusCode))
+	stf.span.Finish()
+}
+
+func parseMethod(req *http.Request) string {
```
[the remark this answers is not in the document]
mikes-blend: Oops these are dead code now. I'll remove them

secrets/vault_transit.go
mikes-blend: We aren't recording them for now, just working on getting 3.0 to GA. Once 3.0 is cut, breaking changes will go on a backlog that will eventually be 4.0

PR #87, opened 2019-06-26
Export more db type fields

alee101:
## PR Summary

Export `Query` fields from `db`

 - **Type:** Internal

#### Reviewers:

### DIFF
--- a/db/invocation.go
+++ b/db/invocation.go
@@ -70,14 +70,14 @@ func (i *Invocation) Query(statement string, args ...interface{}) *Query {
 	var err error
 	statement, err = i.Start(statement)
 	return &Query{
-		context:       i.Context,
-		statement:     statement,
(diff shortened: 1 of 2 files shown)


db/invocation.go
```diff
@@ -70,14 +70,14 @@ func (i *Invocation) Query(statement string, args ...interface{}) *Query {
...
-		conn:          i.Conn,
-		inv:           i,
-		tx:            i.Tx,
-		err:           err,
+		Context:       i.Context,
+		Statement:     statement,
+		CachedPlanKey: i.CachedPlanKey,
+		Args:          args,
+		Conn:          i.Conn,
+		Inv:           i,
```
wcharczuk: with this exported, might make sense to make it fully spelled out: `Invocation`

PR #88, opened 2019-06-28
DB Refactor of the Refactor

wcharczuk:
## PR Summary

This PR unwinds some of the changes Mike made to make it a little less backwards compatibility disrupting, and more consistent within the package but also within the patterns of the repository.

A couple key things:
- Removes the distinction between `Out` and `ReadInto`, and removes `ReadInto`. `Out` will now fully apply any columns seen on a row, potentially setting fields to their zero value if the column contains a null. We do not wipe the object completely on `Out()`.
- `Exec` returns the underlying `sql.Result` from the stdlib function, and new helpers are added (`IgnoreExecResult` and `ExecRowsAffected`) have been added to help use this result.
- Adds `found` as a return on `Query::Scan()` to make it more consistent with `Out` and `ReadInto`.

 - **Type:** Improvements

#### Reviewers:

### DIFF
--- a/db/column.go
+++ b/db/column.go
@@ -131,8 +131,7 @@ func (c Column) SetValue(object interface{}, value interface{}) error {
 			// what should we do here?
 			return ex.New("cannot convert value to destination pointer type")
 		}
-
-		return ex.New("cannot take address of value")
+		return ex.New("cannot take address of value for assignment to object field (which is itself a pointer)")
 	}
 
 	// convert and assign
--- a/db/column.go
+++ b/db/column.go
@@ -69,62 +69,124 @@ func (c Column) SetZero(object interface{}) {
 }
 
 // SetValue sets the field on a database mapped object to the instance of `value`.
-func (c Column) SetValue(object interface{}, value interface{}, resetRowEmpty bool) error {
-	objValue := ReflectValue(object)
-	field := objValue.FieldByName(c.FieldName)
-	fieldType := field.Type()
-	if !field.CanSet() {
-		return ex.New("hit a field we can't set: '" + c.FieldName + "', did you forget to pass the object as a reference?")
+func (c Column) SetValue(object interface{}, value interface{}) error {
+	objectValue := ReflectValue(object)
+
+	objectField := objectValue.FieldByName(c.FieldName)
+	objectFieldType := objectField.Type()
+
+	// check if we've been passed a reference for the object
+	if !objectField.CanSet() {
+		return ex.New("hit a field we can't set; did you forget to pass the object as a reference?").WithMessagef("field: %s", c.FieldName)
 	}
 
 	valueReflected := ReflectValue(value)
 	if !valueReflected.IsValid() {
-		if resetRowEmpty {
-			field.Set(reflect.Zero(field.Type()))
-		}
+		objectField.Set(reflect.Zero(objectFieldType))
 		return nil
 	}
 
+	// special case for `db:"...,json"` fields.
 	if c.IsJSON {
+		// check if we have a driver nullable string for the given value
 		valueContents, ok := valueReflected.Interface().(sql.NullString)
+		// if it's a nullable string, and it's valid (set) and not empty
 		if ok && valueContents.Valid && len(valueContents.String) > 0 {
-			fieldAddr := field.Addr().Interface()
+			// grab a pointer to the field on the object
+			fieldAddr := objectField.Addr().Interface()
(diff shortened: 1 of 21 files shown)


db/connection.go
```diff
@@ -210,15 +210,25 @@ func (dbc *Connection) DefaultSchema() string {
 }
 
 // Exec is a helper stub for .Invoke(...).Exec(...).
-func (dbc *Connection) Exec(statement string, args ...interface{}) (int64, error) {
+func (dbc *Connection) Exec(statement string, args ...interface{}) error {
```
mikes-blend: I still strongly feel that this kind of simplification belongs in a "Manager" for legacy services. Having the default behavior of Exec be blind to its effect is a recipe for writing buggy code.

db/query.go
```diff
@@ -75,66 +77,34 @@ func (q *Query) Scan(args ...interface{}) (err error) {
...
-// or is an "empty" value in the result set, Out clears the field on object it is populating. In short Out maps the
-// output of your query into object as "exactly" as possible. Where you can, prefer Out over Into
-func (q *Query) Out(object interface{}) (found bool, err error) {
-	return q.populateImpl(object, true)
+// ReadInto writes the query result to a single object via. reflection mapping. If there is more than one result, the first
+// result is mapped to to object, and ErrTooManyRows is returned. Unlike Out(), if a field on the stuct is not present
+// or is an "empty" value in the result set, ReadInto() clears the field on object it is populating. In short ReadInto() maps the
+// output of your query into object as "exactly" as possible. Where you can, prefer ReadInto() over Out().
+// If you need multiple queries to fill up your object struct, you should be using Out().
+func (q *Query) ReadInto(object interface{}) (found bool, err error) {
```
mikes-blend: Lets change this name to "OutExclusive" as ReadInto is a better description for what it used to do.

db/connection.go
wcharczuk: it's not, we don't need stats for every call, this preserves existing behavior yes but it's also imho the correct default behavior. manager code that is duplicated for every manager instance should be in the sdk

db/query.go
wcharczuk: ReadInto is used elsewhere in the sdk with this intended behavior (i.e. overwriting completely). It's better for consistency vs. introducing new nomenclature.

mikes-blend: Dislike the name, but consistency is fine.

db/connection.go
mikes-blend: I don't think that managers for new services will duplicate this code (mine certainly won't). I've seen rogue updates/deletes cause issues in connex on several occasions and feel that forcing the issue of acknowledging row count would make developers aware of the fact that they should be checking these things. But I can't make people write correct code in any event.

wcharczuk: https://golang.org/pkg/database/sql/#Result

Actually, you're right; I thought the stdlib exec returned just an error but it's been updated to return a result object. Should we just use that?

mikes-blend: We return the LastUpdateID already (by writing it into the underlying object). That is why I only added the row return count to the method.

wcharczuk: For Exec we don't, it might be good to keep this consistent w/ the stdlib.

mikes-blend: I'm good with that as well.

db/column.go
```diff
@@ -61,71 +61,120 @@ type Column struct {
...
-			if jsonErr != nil {
-				return ex.New(jsonErr)
+		var deserialized interface{}
+		if objectField.Kind() == reflect.Ptr {
+			deserialized = reflect.New(objectField.Type().Elem()).Interface()
+		} else {
+			deserialized = objectField.Addr().Interface()
+		}
+
+		switch valueContents := value.(type) {
```
mikes-blend: NIT: Would it make sense to use reflection to follow all the pointers and get to the underlying type, then switch on that to remove a bunch of these cases?

db/connection.go
```diff
@@ -210,12 +210,12 @@ func (dbc *Connection) DefaultSchema() string {
 }
 
 // Exec is a helper stub for .Invoke(...).Exec(...).
-func (dbc *Connection) Exec(statement string, args ...interface{}) (int64, error) {
+func (dbc *Connection) Exec(statement string, args ...interface{}) (sql.Result, error) {
```
mikes-blend: 👍

db/column.go
wcharczuk: yeah could do, but this way saves a reflect call

PR #89, opened 2019-07-02
Adding support for attaching custom params to an airbrake notice from an error event

softwarebygabe:
## PR Summary

 - **Type:** Features

#### Reviewers:

### DIFF
--- a/airbrake/listener.go
+++ b/airbrake/listener.go
@@ -21,6 +21,8 @@ func AddListeners(log logger.Listenable, cfg Config) {
 	listener := logger.NewErrorEventListener(func(_ context.Context, ee *logger.ErrorEvent) {
 		if req, ok := ee.State.(*http.Request); ok {
 			client.NotifyWithRequest(ee.Err, req)
+		} else if state, ok := ee.State.(Params); ok {
+			client.NotifyWithParams(ee.Err, state)
 		} else {
 			client.Notify(ee.Err)
 		}
--- a/airbrake/listener.go
+++ b/airbrake/listener.go
@@ -21,8 +21,11 @@ func AddListeners(log logger.Listenable, cfg Config) {
 	listener := logger.NewErrorEventListener(func(_ context.Context, ee *logger.ErrorEvent) {
 		if req, ok := ee.State.(*http.Request); ok {
 			client.NotifyWithRequest(ee.Err, req)
-		} else if state, ok := ee.State.(Params); ok {
-			client.NotifyWithParams(ee.Err, state)
+		} else if state, ok := ee.State.(interface{}); ok {
+			errStateMap := map[string]interface{}{
+				"ErrorState": state,
+			}
+			client.NotifyWithParams(ee.Err, errStateMap)
 		} else {
 			client.Notify(ee.Err)
 		}
--- a/airbrake/notice.go
+++ b/airbrake/notice.go
@@ -9,9 +9,6 @@ import (
 	"github.com/blend/go-sdk/webutil"
 )
 
-// Params is a custom type alias
-type Params map[string]interface{}
-
 // NewNotice returns a new gobrake notice.
 func NewNotice(err interface{}, req *http.Request) *gobrake.Notice {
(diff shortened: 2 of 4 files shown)


airbrake/notice.go
```diff
@@ -33,7 +36,7 @@ func NewNotice(err interface{}, req *http.Request) *gobrake.Notice {
 			Context: make(map[string]interface{}),
 			Env:     make(map[string]interface{}),
 			Session: make(map[string]interface{}),
-			Params:  make(map[string]interface{}),
+			Params:  make(Params),
```
wcharczuk: if we do want to keep this as `map string=>interface{}` we'll want to add a field and an option mutator for error events in the logger.

airbrake/notice.go
```diff
@@ -9,6 +9,9 @@ import (
 	"github.com/blend/go-sdk/webutil"
 )
 
+// Params is a custom type alias
+type Params map[string]interface{}
```
wcharczuk: why the type alias?

airbrake/notice.go
```diff
@@ -64,6 +67,13 @@ func NewNotice(err interface{}, req *http.Request) *gobrake.Notice {
 	return notice
 }
 
+// NewNoticeWithParams adds a map of params to the new default gobrake notice.
+func NewNoticeWithParams(err interface{}, params map[string]interface{}) *gobrake.Notice {
```
dhermes: `map[string]interface{}` or your `Params` alias here?

softwarebygabe: good catch, got rid of the alias now

airbrake/listener.go
```diff
@@ -21,6 +21,11 @@ func AddListeners(log logger.Listenable, cfg Config) {
 	listener := logger.NewErrorEventListener(func(_ context.Context, ee *logger.ErrorEvent) {
 		if req, ok := ee.State.(*http.Request); ok {
 			client.NotifyWithRequest(ee.Err, req)
+		} else if state, ok := ee.State.(interface{}); ok {
+			errStateMap := map[string]interface{}{
+				"ErrorState": state,
+			}
```
softwarebygabe: how do we feel about this approach?

airbrake/notice.go
softwarebygabe: rm'd

airbrake/listener.go
softwarebygabe: basically now just passing anything on the error state unless it is of the specific http request type

softwarebygabe:
Now the API, if you wanted to airbrake custom params from the logger would look like:
```
// add airbrake listeners
if !cfg.Airbrake.IsZero() {
  airbrake.AddListeners(log, cfg.Airbrake)
}

// airbrake something
log.ErrorWithState(errors.New("destined for airbrake"), map[string]string{
  "some_relevant_key": "the_value",
})
```

airbrake/listener.go
wcharczuk: to put it bluntly, the type assertion `.(interface{})` is pointless, everything ascribes to the empty interface, just test if it's != nil

logger/context.go
```diff
@@ -142,14 +142,18 @@ func (sc Context) Error(err error) error {
 	return err
 }
 
-// ErrorWithReq logs an error to std err with a request.
-func (sc Context) ErrorWithReq(err error, req *http.Request) error {
-	ee := NewErrorEvent(Error, err)
-	ee.State = req
+// ErrorWithState logs an error to std with a generic state.
+func (sc Context) ErrorWithState(err error, state interface{}) error {
```
wcharczuk: i wouldn't have both this and ErrorWithReq, i would have an `...ErrorOption` variadic param on `Error()`, `Warning()` and `Fatal()` that mutates the error event object, setting the state, or setting a request property.

wcharczuk:
@softwarebygabe hey we're going to close this out in favor of the sentry tooling we're going to add soon. will circle back with you on this after.

PR #90, opened 2019-07-16
Add websocket support and tests

JohnlNguyen:
## PR Summary

 - **Type:** Improvements

#### Reviewers:


proxy/upstream.go
```diff
@@ -49,6 +49,9 @@ func (u *Upstream) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	start := time.Now()
 
 	w := NewResponseWriter(rw)
+	if hijacker, ok := w.innerResponse.(http.Hijacker); ok {
```
chotiwat: While this works, it's not probably a good pattern because the ResponseWriter doesn't encapsulate the hijacker logic. You could forget to assign an innerResponse as `w.Hijacker` and get nil ref when `w.Hijack()` is called.

I think a better way to do this is to implement a `Hijack()` function for `ResponseWriter` which internally calls `Hijack` of the `innerResponse`.

proxy/proxy_test.go
```diff
@@ -1,10 +1,17 @@
 package proxy
 
 import (
+	"bufio"
+	"fmt"
+	"github.com/gorilla/websocket"
+	"golang.org/x/net/http/httpguts"
```
chotiwat: please move all non standard packages to the second block

wcharczuk:
can you make sure to apply this to master in addition to v2.0 if it still applies?

proxy/response_writer.go
```diff
@@ -30,6 +34,18 @@ func (rw *ResponseWriter) Header() http.Header {
 	return rw.innerResponse.Header()
 }
 
+// Hijack assign hijacker to inner response if null and calls Hijack of inner response
+func (rw *ResponseWriter) Hijack() (net.Conn, *bufio.ReadWriter, error) {
+	if rw.hijacker == nil {
+		hijacker, ok := rw.innerResponse.(http.Hijacker)
+		if !ok {
+			return nil, nil, fmt.Errorf("can't assign Hijacker to innerResponse")
```
JohnlNguyen: not sure what is the best way to handle this case when `!ok`, I'm open to suggestions here

chotiwat: yeah, I think this looks good. it's what this guy does as well https://github.com/go-martini/martini/pull/64/files#diff-522a3ef72074d4d132ff8627e125f3b3R61

I'd also check other go-sdk packages if they return vanilla errors or exception-wrapped ones.

wcharczuk: where possible, use `ex.New(ErrClass).WithMessage()` etc.

proxy/response_writer.go
```diff
@@ -30,6 +34,18 @@ func (rw *ResponseWriter) Header() http.Header {
...
+		}
+		rw.hijacker = hijacker
```
chotiwat: will `Hijack()` be called more than once? if not, we don't need to cache it.

@wcharczuk any insight on this?

wcharczuk: i doubt it would be called more than once, but what you can do is panic if that happens (i.e. invert this check)

wcharczuk: fwiw either approach is bad for different reasons; panics are generally catastrophic and hard to debug, having a set if unset pattern makes reasoning about the call harder b/c you may not know why nothing happened after the call

proxy/response_writer.go
```diff
@@ -30,6 +34,18 @@ func (rw *ResponseWriter) Header() http.Header {
 	return rw.innerResponse.Header()
 }
 
+// Hijack assign hijacker to inner response if null and calls Hijack of inner response
+func (rw *ResponseWriter) Hijack() (net.Conn, *bufio.ReadWriter, error) {
```
wcharczuk: it is unclear from the docs what this does? what does it mean to hijack the response?

proxy/response_writer.go
```diff
@@ -14,6 +17,7 @@ func NewResponseWriter(w http.ResponseWriter) *ResponseWriter {
 // ResponseWriter a better response writer
 type ResponseWriter struct {
 	innerResponse http.ResponseWriter
+	hijacker http.Hijacker
```
wcharczuk: do we need to keep this reference around? can we scope it to the call of Hijack()?

PR #92, opened 2019-07-19
Adds `validate` package

wcharczuk:
## PR Summary

This PR adds a `validate` package that's meant to mirror the useful parts of JOI. 

It make slight ergonomic tweaks to the `ex` package to allow it to be used for validation errors.

 - **Type:** Features

#### Reviewers:


validate/string.go
```diff
@@ -0,0 +1,131 @@
...
+			return ex.New(err)
+		}
+		if !exp.MatchString(string(s)) {
+			return Errorf(ErrStringMatches, "expression: %s", expression)
+		}
+		return nil
+	}
+}
+
+// IsUpper returns a validator if a string is all uppercase.
```
alee101: same for these, either update comment or return a validator?

validate/time.go
```diff
@@ -0,0 +1,67 @@
...
+func (t Time) Before(before time.Time) Validator {
+	return func() error {
+		if time.Time(t).After(before) {
+			return Errorf(ErrTimeBefore, "before: %v", before)
+		}
+		return nil
+	}
+}
+
+// BeforeNowUTC returns a validator that a time should be before a given time.
```
alee101: same as above

validate/validated.go
```diff
@@ -0,0 +1,21 @@
+package validate
+
+// Validated is a root interface for validated types.
+type Validated interface {
```
alee101: maybe call this `Validateable`? `IsValidated` suggests there might be a difference between the pre-validation and post-validation state

validate/all.go
```diff
@@ -0,0 +1,13 @@
+package validate
+
+// All returns the all the failing validations.
```
alee101: `returns all the`

validate/any.go
```diff
@@ -0,0 +1,130 @@
...
+func Any(obj interface{}) AnyValidators {
+	return AnyValidators{Obj: obj}
+}
+
+// AnyValidators are validators for the empty interface{}.
+type AnyValidators struct {
+	Obj interface{}
+}
+
+// Zero retruns a validator that asserts an object is it's zero value.
```
alee101: `returns`

validate/int.go
```diff
@@ -0,0 +1,68 @@
...
+		}
+		if int(i) > max {
+			return Errorf(ErrIntMax, "max: %d", max)
+		}
+		return nil
+	}
+}
+
+// Positive returns a validator that an int is positive.
+func (i IntValidators) Positive() error {
```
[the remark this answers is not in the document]
wcharczuk: done

validate/string.go
wcharczuk: done

validate/time.go
wcharczuk: done

validate/validated.go
wcharczuk: IsValidated is a type assertion that makes sure an object implements a `Validate() error` function, it doesn't connote that function has been called. I should probably make that clear in the docs.

validate/all.go
wcharczuk: done

validate/any.go
wcharczuk: done

validate/int.go
```diff
@@ -24,7 +24,7 @@ type IntValidators struct {
 func (i IntValidators) Min(min int) Validator {
 	return func() error {
 		if i.Value == nil || *i.Value < min {
-			return Errorf(ErrIntMin, "min: %d", min)
+			return Errorf(ErrIntMin, i.Value, "min: %d", min)
```
alee101: dereference value here and below

wcharczuk: fixed

PR #93, opened 2019-07-21
Fixes issues with the int validators with nil inputs, adds float64 validators

wcharczuk:
## PR Summary

Fixes issues with the int validators with nil inputs, and adds float64 validators.

 - **Type:** Bugfix

#### Reviewers:

### DIFF
--- a/validate/float64.go
+++ b/validate/float64.go
@@ -1,0 +1,121 @@
+package validate
+
+import (
+	"math"
+
+	"github.com/blend/go-sdk/ex"
+)
+
+// Int errors
+const (
+	ErrFloat64Min      ex.Class = "float64 should be above a minimum value"
+	ErrFloat64Max      ex.Class = "float64 should be below a maximum value"
+	ErrFloat64Positive ex.Class = "float64 should be positive"
+	ErrFloat64Negative ex.Class = "float64 should be negative"
+	ErrFloat64Epsilon  ex.Class = "float64 should be within an epsilon of a value"
+)
+
+// Float64 returns validators for float64s.
+func Float64(value *float64) Float64Validators {
+	return Float64Validators{value}
+}
+
+// Float64Validators implements float64 validators.
+type Float64Validators struct {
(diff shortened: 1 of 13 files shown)


validate/string.go
```diff
@@ -97,7 +97,8 @@ func (s StringValidators) BetweenLen(min, max int) Validator {
 }
 
 // Matches returns a validator that a string matches a given regex.
-// If the value is unset it will match, i.e. not fail validation.
+// If the value is unset (nil) it will match, i.e. not fail validation and no error will
```
mikes-blend: Is this behavior desirable? I would think that we'd want to fail this on a nil value?

mikes-blend: Is this behavior desirable? I'd think it would be better to fail this on a nil.

validate/float64.go
```diff
@@ -0,0 +1,121 @@
...
+const (
+	ErrFloat64Min      ex.Class = "float64 should be above a minimum value"
+	ErrFloat64Max      ex.Class = "float64 should be below a maximum value"
+	ErrFloat64Positive ex.Class = "float64 should be positive"
+	ErrFloat64Negative ex.Class = "float64 should be negative"
+	ErrFloat64Epsilon  ex.Class = "float64 should be within an epsilon of a value"
+)
+
+// Float64 returns validators for float64s.
+func Float64(value *float64) Float64Validators {
```
mikes-blend: Should we provide these for other fixed size numeric types?

validate/string.go
wcharczuk: i wanted to mirror the intent of joi, strictly speaking i'd want this to fail but i also want to be able to port validators easily, lemme dig in some more

validate/float64.go
wcharczuk: generally i've picked the most oft used ones because doing all of them is quite heavy weight (uint8 -> uint64, int8 -> int64, float32 etc.) we can add more over time as we see needs?

PR #95, opened 2019-07-29
Refactors web.AuthManager Constructor to use the options pattern

wcharczuk:
## PR Summary

Refactors web.AuthManager Constructor to use the options pattern.

As a result, the top level `web.New()` constructor will now also return an error. A `MustNew()` form has been added for convenience and to smooth refactoring.

It also adds `.Header` to the response logger event metadata for use in tactical situations where you want to test for things in the headers. 

Some other smaller tweaks:
- removes the logger reference from `web.Ctx`
- changes slightly a few function names on logger metadata maps
- addresses inconsistencies with web.ResultProvider implementations returning 401 vs. 403 (initiated previous to this PR).

 - **Type:** Internal

#### Reviewers:


web/auth_manager.go
```diff
@@ -9,76 +9,189 @@ import (
...
-	manager.CookiePath = cfg.CookiePathOrDefault()
-	manager.CookieSameSite = cfg.CookieSameSiteOrDefault()
-	manager.SessionTimeoutProvider = SessionTimeoutProvider(!cfg.SessionTimeoutIsRelative, cfg.SessionTimeoutOrDefault())
-	return manager
+func NewAuthManager(options ...AuthManagerOption) (manager AuthManager, err error) {
+	manager.CookieDefaults.Name = DefaultCookieName
+	manager.CookieDefaults.Secure = DefaultCookieSecure
+	manager.CookieDefaults.HttpOnly = DefaultCookieHTTPOnly
+	manager.CookieDefaults.SameSite = DefaultCookieSameSite
+
```
alee101: should Path also be defaulted here?

web/ctx.go
```diff
@@ -30,22 +29,47 @@ func NewCtx(w ResponseWriter, r *http.Request, options ...CtxOption) *Ctx {
...
-	Views           *ViewCache
-	Log             logger.Log
-	Tracer          Tracer
-	Body            []byte
-	Form            url.Values
+	// App is a reference back to the parent application.
+	App *App
+	// Auth is a reference to the app default auth manager, but
+	// can be overwritten by middleware.
+	Auth AuthManager
```
alee101: for what cases should the context level `ctx.Auth` be used over `ctx.App.Auth`?

web/auth_manager.go
wcharczuk: yep, 100%

web/ctx.go
wcharczuk: Thinking is you can override the app default with middleware on a per route / controller basis if you want to switch how different "areas" of an app authenticate.

web/auth_manager.go
wcharczuk: done

PR #96, opened 2019-07-30
Don't set the forwarded headers on the response

mat285:
## PR Summary

Remove x-forwarded-for and x-forwarded-port from the reverseproxy response

 - **Type:** Bugfix
 - **Issue Link:** 

#### Reviewers:

### DIFF
--- a/reverseproxy/upstream_test.go
+++ b/reverseproxy/upstream_test.go
@@ -1,6 +1,9 @@
 package reverseproxy
 
 import (
+	"fmt"
+	"net/http"
+	"net/http/httptest"
 	"testing"
 
 	"github.com/blend/go-sdk/assert"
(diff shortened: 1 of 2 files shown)


mat285:
Requires a backport

reverseproxy/upstream_test.go
```diff
@@ -27,7 +28,7 @@ func TestUpstreamServeHTTP(t *testing.T) {
 
 	u := NewUpstream(MustParseURL(srv.URL))
 	assert.NotNil(u.ReverseProxy)
-	u.Log = nil
+	u.Log = logger.None()
```
wcharczuk: we should make sure Upstream can operate w/o a logger, can you look where it's using the logger and add wrapping `logger.Maybe...` calls?

mat285: I only added this to bump the test coverage. It doesn't need a logger don't worry

mat285:
bailey merge it

reverseproxy/upstream_test.go
wcharczuk: this is going into my "why coverage minimums are stupid" folder

PR #99, opened 2019-08-03
OAuth tracing improvments

wcharczuk:
## PR Summary

Adds a number of fixes to web and a couple other packages:
- Switches `StaticFileServer` to use the options pattern
- Fixes issues with the static file server nil-refing on cached mode not found.
- Improves performance of static file server in high throughput scenarios by switching to a read / write mutex for the map.
- Tweaks how the auth manager treats cookie defaults, specifically for the `SameSite` parameter.
- Adds a `oauthtrace` implementation for the `oauth` tracer.
- Tweaks the `r2trace` to add more information to traces.

 - **Type:** Improvements

#### Reviewers:


web/static_file_server.go
```diff
@@ -124,9 +186,22 @@ func (sc *StaticFileServer) ResolveFile(filePath string) (f http.File, err error
...
 // It returns the cached instance of a file if it exists, and adds it to the cache if there is a miss.
 func (sc *StaticFileServer) ResolveCachedFile(filepath string) (*CachedStaticFile, error) {
+	sc.RLock()
+	if sc.Cache != nil {
+		if file, ok := sc.Cache[filepath]; ok {
+			sc.RUnlock()
+			return file, nil
+		}
+	}
+	sc.RUnlock()
```
gcarling: could wrap the above in a func and use defer if you want, dunno the value of making sure something doesn't crash and cause a deadlock here is

web/static_file_server.go
```diff
@@ -124,9 +186,22 @@ func (sc *StaticFileServer) ResolveFile(filePath string) (f http.File, err error
...
+
 	sc.Lock()
 	defer sc.Unlock()
 
+	if sc.Cache == nil {
```
gcarling: could move this to the top of this function and then skip the `if sc.Cache != nil` above

gcarling:
`Intended Change Level: major` seems accurate

cmd/profanity/main.go
```diff
@@ -92,22 +95,23 @@ func main() {
 	cmd.Run = func(parent *cobra.Command, args []string) {
 		var cfg config
 		if err := cfg.Resolve(); err != nil {
-			fmt.Fprintf(os.Stderr, "%+v\n\n", err)
```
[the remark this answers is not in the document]
wcharczuk: yes, it removes the stack from the output which is annoying af in profanity errors

ex/util.go
```diff
@@ -40,6 +40,21 @@ func ErrClass(err interface{}) error {
 	return nil
 }
 
+// ErrStackTrace returns the exception stack trace.
+// This depends on if the err is itself an exception or not.
+func ErrStackTrace(err interface{}) StackTrace {
```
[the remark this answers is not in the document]
wcharczuk: It's just adding it for later / other use cases. It mirrors the other `Err...` functions that extract fields from errors if they're exceptions safely

logger/write_helpers.go
```diff
@@ -20,8 +20,13 @@ func WriteHTTPRequest(tf TextFormatter, wr io.Writer, req *http.Request) {
 		io.WriteString(wr, Space)
 	}
 	io.WriteString(wr, tf.Colorize(req.Method, ansi.ColorBlue))
-	io.WriteString(wr, Space)
-	io.WriteString(wr, req.URL.Path)
+	if req.URL.RawPath != "" {
```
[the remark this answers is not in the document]
wcharczuk: the old behavior is it would write the space, this will now omit the space.

The raw path is desirable because it contains the query string if relevant.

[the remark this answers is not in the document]
wcharczuk: i've changed this to just string the url

web/static_file_server.go
wcharczuk: https://en.wikipedia.org/wiki/Double-checked_locking

wcharczuk: basically, while you were waiting to acquire that lock, another thread might have set up the cache and you don't need to do the init anymore

gcarling: ah nice

PR #101, opened 2019-08-09
Listen for TLS connections before accepting proxy protocol connections.

alexfish8:
## PR Summary

When setting up a web service to use a listener created by `proxyprotocol.CreateListener` (a listener configured to handle proxyprotocol and TLS), I was getting TLS handshake errors when attempting to talk to the service behind an AWS classic ELB. Switching the order of these listeners fixes this TLS handshake issue.

 - **Type:** Bugfixes

#### Reviewers:

### DIFF
--- a/proxyprotocol/create_listener_test.go
+++ b/proxyprotocol/create_listener_test.go
@@ -1,6 +1,8 @@
 package proxyprotocol
 
 import (
+	"crypto/tls"
+	"reflect"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


proxyprotocol/create_listener_test.go
```diff
@@ -31,3 +33,22 @@ func TestCreateListener(t *testing.T) {
...
+		OptKeepAlivePeriod(30*time.Second),
+		OptTLSConfig(tlsConfig),
+	)
+	defer listener.Close()
+
+	assert.Nil(err)
+	assert.NotNil(listener)
+
+	// hacky use of reflection to verify that the created listener can handle tls connections
+	assert.Equal("*tls.listener", reflect.TypeOf(listener).String())
```
wcharczuk: heads up, you can do a type assertion here instead of a strings compare

```_, isTyped := listener.(*tls.Listener); assert.True(isTyped)```

alexfish8: `tls.listener` is private, so the type assertion won't work. I'm trying to think of a way to test this that's more behavioral

wcharczuk: that's super random, does tls export an interface type that listener ascribes to?

wcharczuk: because then you could assert the listener implements that interface but it's not really strict enough problably

alexfish8: Looks like `tls.Conn` is public, so I'll update the test to verify the type of an accepted connection instead.

alexfish8:
@wcharczuk It doesn't look like this helper function exists in `v2.0`. Do you want me to just backport the helper?

PR #102, opened 2019-08-10
Adds `breaker` package.

wcharczuk:
## PR Summary

Adds a simple circuit breaker package. It is designed with making calls to flaky resources in mind, but does not constrain the caller to http calls (i.e. can be used to make any type of call that can be flaky).

 - **Type:** Features

#### Reviewers:

### DIFF
--- a/breaker/breaker.go
+++ b/breaker/breaker.go
@@ -14,8 +14,8 @@ type (
 	Action func(context.Context) (interface{}, error)
 	// OnStateChangeHandler is called when the state changes.
 	OnStateChangeHandler func(ctx context.Context, from, to State, generation int64)
-	// ShouldCloseProvider returns if the breaker should close.
-	ShouldCloseProvider func(ctx context.Context, counts Counts) bool
+	// ShouldOpenProvider returns if the breaker should open.
+	ShouldOpenProvider func(ctx context.Context, counts Counts) bool
 	// NowProvider returns the current time.
 	NowProvider func() time.Time
 )
@@ -54,10 +54,10 @@ type Breaker struct {
 
 	// OnStateChange is an optional handler called when the breaker transitions state.
 	OnStateChange OnStateChangeHandler
-	// ShouldCloseProvider is called optionally to determine if we should close the breaker.
-	ShouldCloseProvider ShouldCloseProvider
+	// ShouldOpenProvider is called optionally to determine if we should open the breaker.
+	ShouldOpenProvider ShouldOpenProvider
 	// NowProvider lets you optionally inject the current time for testing.
-	NowProvider
+	NowProvider NowProvider
 
 	// HalfOpenMaxActions is the maximum number of requests
 	// we can make when the state is HalfOpen.
@@ -69,7 +69,6 @@ type Breaker struct {
 	// after which the state of the CircuitBreaker becomes half-open.
 	// If Timeout is 0, the timeout value of the CircuitBreaker is set to 60 seconds.
 	OpenExpiryInterval time.Duration
-	// State is the current breaker state (Open, HalfOpen, Closed).
 	// Counts are stats for the breaker.
 	Counts Counts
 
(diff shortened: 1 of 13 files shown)


breaker/breaker.go
```diff
@@ -0,0 +1,241 @@
...
+	// we can make when the state is HalfOpen.
+	HalfOpenMaxActions int64
+	// ClosedExpiryInterval is the cyclic period of the closed state for the CircuitBreaker to clear the internal Counts.
+	// If Interval is 0, the CircuitBreaker doesn't clear internal Counts during the closed state.
+	ClosedExpiryInterval time.Duration
+	// OpenExpiryInterval is the period of the open state,
+	// after which the state of the CircuitBreaker becomes half-open.
+	// If Timeout is 0, the timeout value of the CircuitBreaker is set to 60 seconds.
+	OpenExpiryInterval time.Duration
+	// State is the current breaker state (Open, HalfOpen, Closed).
```
mikes-blend: Typo.. State comment is below

breaker/breaker.go
```diff
@@ -0,0 +1,241 @@
...
+	return res, err
+}
+
+// State returns the current state of the CircuitBreaker.
+func (b *Breaker) State(ctx context.Context) State {
+	b.Lock()
+	defer b.Unlock()
+
+	now := time.Now()
+	state, _ := b.getState(ctx, now)
```
mikes-blend: This increments counters and potentially toggles breaker state in getState, which is probably not the desired action for an exported state check.

breaker/breaker.go
```diff
@@ -0,0 +1,241 @@
...
+
+// Do runs the given action if the Breaker accepts it.
+// Do returns an error instantly if the Breaker rejects the request.
+// Otherwise, Do returns the result of the request.
+// If a panic occurs in the request, the Breaker handles it as an error.
+func (b *Breaker) Do(ctx context.Context, action Action) (interface{}, error) {
+	generation, err := b.beforeAction(ctx)
+	if err != nil {
+		if b.OpenAction != nil {
+			return b.OpenAction(ctx)
```
mikes-blend: Do there need to be metrics updates around this? Should we be tracking "OpenActions" taken?

breaker/breaker.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+}
+
+func (b *Breaker) failure(ctx context.Context, state State, now time.Time) {
+	switch state {
+	case StateClosed:
+		atomic.AddInt64(&b.Counts.TotalFailures, 1)
+		atomic.AddInt64(&b.Counts.ConsecutiveFailures, 1)
+		atomic.StoreInt64(&b.Counts.ConsecutiveSuccesses, 0)
+		if b.shouldClose(ctx) {
```
mikes-blend: shouldOpen? This doesn't seem to make logical sense. If I'm in the closed state, why am I checking shouldClose to see if I should Open?

breaker/breaker.go
```diff
@@ -0,0 +1,241 @@
...
+	ShouldCloseProvider ShouldCloseProvider
+	// NowProvider lets you optionally inject the current time for testing.
+	NowProvider
+
+	// HalfOpenMaxActions is the maximum number of requests
+	// we can make when the state is HalfOpen.
+	HalfOpenMaxActions int64
+	// ClosedExpiryInterval is the cyclic period of the closed state for the CircuitBreaker to clear the internal Counts.
+	// If Interval is 0, the CircuitBreaker doesn't clear internal Counts during the closed state.
+	ClosedExpiryInterval time.Duration
```
mikes-blend: I'm not understanding the purpose of this. Can you elaborate on why we need to clear these counts? Breaker toggle appears to be based on successes/failures.

wcharczuk: if you enter a `Open` state, and some time passes, you enter a `HalfOpen` state where you are allowing requests but if there is another failure immediately `Open` fully. this mirrors the origin package handling of this.

breaker/breaker.go
wcharczuk: this is handled within `beforeAction`

breaker/breaker.go
wcharczuk: because state is evaluated rather than strictly returned, we may want to rename this `EvaluateState`

breaker/breaker.go
wcharczuk: open vs. closed hurts my brain.

breaker/breaker.go
mikes-blend: I get the HalfOpenMaxActions part, just not the ClosedExpiryInterval.

PR #103, opened 2019-08-13
Add multi-write stats collector 

JohnlNguyen:
## PR Summary

Adding a multi-write stats collector to write to multiple stats-D hosts. It wraps a set of datadog collectors.

* **Type:** Features
* **Intended Change Level:** minor

#### Reviewers:

Does the "Intended Change Level" above match the actual behavior of this PR?
Is this PR following patterns set forth in the package (if modifying a package), or the go-sdk design patterns if implementing a new package from scratch?
Does this require a backport to LTS versions? If so ping, let the contributors group know.

### DIFF
--- a/stats/mock_collector.go
+++ b/stats/mock_collector.go
@@ -16,7 +16,7 @@ var (
 func NewMockCollector() *MockCollector {
 	return &MockCollector{
 		Events: make(chan MockMetric),
-		Errors: make(chan error, 30),
+		Errors: make(chan error),
 	}
 }
 
--- a/stats/mock_collector.go
+++ b/stats/mock_collector.go
@@ -16,6 +16,7 @@ var (
 func NewMockCollector() *MockCollector {
 	return &MockCollector{
 		Events: make(chan MockMetric),
+		Errors: make(chan error, 30),
 	}
 }
 
@@ -25,6 +26,7 @@ type MockCollector struct {
 	defaultTags []string
 
 	Events chan MockMetric
+	Errors chan error
 }
 
 // AddDefaultTag adds a default tag.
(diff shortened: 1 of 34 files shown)


JohnlNguyen:
@wcharczuk @chotiwat can you guys take a look

stats/multicollector.go
```diff
@@ -0,0 +1,14 @@
+package stats
+
+import (
+	"time"
+)
+
+// MultiCollector is a multi-write stats collector
+type MultiCollector interface {
```
wcharczuk: why can't we re-use the existing interface and just swap out the type that implements it?

stats/stats_collector.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/blend/go-sdk/ex"
+)
+
+// type assertion that StatsCollector implements MultiCollector
+var (
+	_ MultiCollector = (*StatsCollector)(nil)
+)
+
+// StatsCollector is a class that wraps a set of statsd collectors
+type StatsCollector struct {
```
wcharczuk: I would rename this `MultiCollector` and unless it needs to have extra state on it, just type alias `type MultiCollector []Collector`

stats/multicollector.go
JohnlNguyen: done

stats/stats_collector.go
JohnlNguyen: done

stats/multicollector.go
```diff
@@ -0,0 +1,68 @@
...
+	"time"
+
+	"github.com/blend/go-sdk/ex"
+)
+
+// MultiCollector is a class that wraps a set of statsd collectors
+type MultiCollector []Collector
+
+// NewMultiCollector returns a new multi-collector from a set of collectors
+func NewMultiCollector(collectors ...Collector) (MultiCollector, error) {
```
wcharczuk: i honestly wouldn't worry about this constructor; the `range` operator will handle nil or empty inputs fine

wcharczuk:
looks like there is still untested code that codecov is complaining about; probably need to add more tests. you can look at the codecov report or the `coverage.html` that is generated by running `make cover` locally

stats/mock_collector.go
```diff
@@ -16,6 +16,7 @@ var (
 func NewMockCollector() *MockCollector {
 	return &MockCollector{
 		Events: make(chan MockMetric),
+		Errors: make(chan error, 30),
```
wcharczuk: if the `Events` chan does't have a length the errors chan probably shouldn't either.

JohnlNguyen: done

PR #105, opened 2019-08-15
Wcharczuk/logger scope

wcharczuk:
## PR Summary

Refactors the logger scope, additionally refactors how event meta is propagated and what is held on the events themselves vs. what is stashed on the context. This causes some breaking changes. 

It also removes airbrake from a couple internal utilities, but keeps the airbrake package.

Also adds webutil tests for request options.

 - **Type:** Improvements|

#### Reviewers:


logger/http_request_event.go
```diff
@@ -2,7 +2,6 @@ package logger
```
mikes-blend: Does this file belong in package web or r2?

logger/event.go
```diff
@@ -1,24 +1,27 @@
...
 import (
-	"context"
+	"io"
 	"time"
 )
 
 // Event is an interface representing methods necessary to trigger listeners.
 type Event interface {
 	GetFlag() string
-	GetTimestamp() time.Time
```
mikes-blend: Why was timestamp removed? I'd think that all events would have a Timestamp?

logger/http_response_event.go
```diff
@@ -2,7 +2,6 @@ package logger
```
mikes-blend: Again is this the right package for this event?

logger/http_request_event.go
wcharczuk: it's used by `webutil`, `web`. `r2` has it's own client request event.

logger/event.go
wcharczuk: one fewer thing to worry about in event metadata, we can add back though

wcharczuk: it's also not strictly required

logger/http_response_event.go
wcharczuk: with grpc coming, it might make more sense to move to webutil

logger/event.go
mikes-blend: Don't have super strong feelings here, but could see it being useful.

wcharczuk: yeah, we also get into situations where the log trigger queue gets long and there is meaningful skew between when an event is triggered and when the handler is called, so it's helpful as part of the event

mikes-blend: 👍 time-of-occurrence vs time-of-handling distinction is nice to be able to track

examples/logger/meta/main.go
```diff
@@ -0,0 +1,13 @@
+package main
+
+import "github.com/blend/go-sdk/logger"
+
+// L is a helper type alias.
+type L = logger.Labels
```
gcarling: i realize its fewer characters but seems more likely to cause confusion than help imo

logger/context.go
```diff
@@ -2,164 +2,73 @@ package logger
...
+func WithAnnotations(ctx context.Context, annotations Annotations) context.Context {
+	return context.WithValue(ctx, annotationsKey{}, annotations)
 }
 
-// FatalWithReq logs an error as fatal with a request as state.
-func (sc Context) FatalWithReq(err error, req *http.Request) error {
-	ee := NewErrorEvent(Fatal, err)
-	ee.State = req
-	sc.Trigger(context.Background(), ee)
-	return err
```
gcarling: what is left in this file? seems like its almost entirely deleted - seems odd to keep around the logger specific context along with the new `Scope` concept

logger/event.go
```diff
@@ -2,23 +2,41 @@ package logger
 
 import (
 	"context"
+	"io"
 	"time"
 )
 
 // Event is an interface representing methods necessary to trigger listeners.
 type Event interface {
```
gcarling: does the concept of an `event` here tie nicely into the concept of an events w.r.t the events pipeline?

logger/logger_interfaces.go
```diff
@@ -12,10 +12,12 @@ type Triggerable interface {
 	Trigger(context.Context, Event)
```
gcarling: new name on this file feels odd. seems like `logger/logger_interfaces` is kind of repetitive

logger/logger_interfaces.go
```diff
@@ -12,10 +12,12 @@ type Triggerable interface {
...
 }
 
-// SubContexter is a type that can return a subcontext.
-type SubContexter interface {
-	SubContext(string, ...ContextOption) Context
-	WithFields(Fields, ...ContextOption) Context
+// Scoper is a type that can return a scope.
+type Scoper interface {
```
gcarling: one odd thing about Scope vs. SubContext is the former is kind of unclear to me how it differs from context, whereas the latter is more obviously a more specific context for this scenario. 

i'm curious how you are thinking about the difference in *functionality* change vs. the difference in the *name* change - could we be making this same change but still calling it a `SubContext`?

logger/scope.go
```diff
@@ -0,0 +1,191 @@
...
+// OptScopeContext sets the context on a scope.
+// This context will be used as the triggering context for any events.
+func OptScopeContext(ctx context.Context) ScopeOption {
+	return func(s *Scope) {
+		s.Context = ctx
+	}
+}
+
+// WithContext returns a new scope context.
+func (sc Scope) WithContext(ctx context.Context) Scope {
```
gcarling: this is pretty interesting - in the past `WithX` has mutated and returned, rather than returning a *new* struct with the existing options. what's motivating the shift here

wcharczuk: copies are easier to reason about vs. mutating the original object. the benefit here is if you return a `With...` copy here, and do another `With...` later in a different branch, they won't interfere with each other.

logger/logger_interfaces.go
wcharczuk: Functionally they're the same, the issue was anything resembling `context.Context` was confusing.

logger/logger_interfaces.go
wcharczuk: yeah it's tricky because we have event interfaces, and logger interfaces, i agree it stutters but it's ok

logger/event.go
wcharczuk: that's the goal; and why the scope has `Labels` and `Annotations` and you can add labels and annotations to the trigger context

logger/context.go
wcharczuk: the `context` utilities in this file are for you to add labels and annotations to the trigger context, in addition to the re-usable labels and annotations on the scope

examples/logger/meta/main.go
wcharczuk: it's in an example, i do this elsewhere too but would never put it in the library itself.

PR #107, opened 2019-08-20
Assert IndirectEqual

akarimcheese:
## PR Summary

Adds an assert function for checking equality pointers if nil, and if not then checks the equality of the values that the pointers point to

 - **Type:** Improvements

#### Reviewers:

### DIFF
--- a/assert/assert.go
+++ b/assert/assert.go
@@ -172,6 +172,14 @@ func (a *Assertions) ReferenceEqual(expected interface{}, actual interface{}, us
 	}
 }
 
+// IndirectEqual asserts that two pointers reference the same values.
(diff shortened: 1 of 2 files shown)


akarimcheese:
open to changing the name of the assert function

assert/assert.go
```diff
@@ -1210,6 +1243,17 @@ func areReferenceEqual(expected, actual interface{}) bool {
 	return expected == actual
 }
 
+func areIndirectEqual(expected, actual interface{}) bool {
+	if expected == nil && actual == nil {
```
akarimcheese: This check doesn't work if you pass in typed nils...

wcharczuk:
can you verify `reflect.DeepEqual` doesn't do what you want it to do here?

akarimcheese:
It does, as does `assert.Equal` 🤦‍♂ . ty!

PR #110, opened 2019-08-27
Serialization and deserialization for environment variables

afnanenayet:
## PR Summary

Due to some features that have been requested in projects that depend on go-sdk, there is a need to have some sort of method that serializes and deserializes strings to environment variables. This PR introduces a serialization and deserialization method with a well-defined grammar.

I'm not super tied to the names, so let me know if you think the functions could have better names.

 - **Type:** Features

#### Reviewers:


afnanenayet:
> One last thing: we should rename `split_semi.go` to something a bit more descriptive (maybe like `parse.go`, or `string.go`, but even these names are a bit vague).

done

PR #117, opened 2019-09-04
Backport of assert fixes

wcharczuk:
This applies the master fixes for assert to v2


COVERAGE
```diff
@@ -1 +1 @@
-42.66
```
gcarling: :(

PR #121, opened 2019-09-13
Adds `bindata` utility for adding static assets to binaries

wcharczuk:
## PR Summary

Adds the `bindata` command-line utility and library package.

`bindata` generates a single `.go` file for static assets to be fetched from memory, allowing you to pack things like views into a cli program. It does this by gzipping the static files it discovers in a directory, recording extra metadata like the modtime.

 - **Type:** Features

#### Reviewers:


bindata/bundle_test.go
```diff
@@ -0,0 +1,32 @@
...
+	err := bundle.Process(buffer, PathConfig{Path: "./testdata", Recursive: true})
+	assert.Nil(err)
+
+	assert.NotEmpty(buffer.Bytes())
+
+	assert.Contains(buffer.String(), "package bindata")
+	assert.Contains(buffer.String(), "testdata/js/app.js")
+	assert.Contains(buffer.String(), "testdata/css/site.css")
+
+	ast, err := parser.ParseFile(token.NewFileSet(), "bindata.go", buffer.Bytes(), parser.ParseComments|parser.AllErrors)
```
akarimcheese: a general q, not related to the quality of this test:
for code gen, is it ever reasonable generate asts and pretty print them into source files?

bindata/bundle_test.go
```diff
@@ -0,0 +1,32 @@
...
+	assert.Nil(err)
+	assert.NotNil(ast)
+	assert.Len(ast.Imports, 5)
```
akarimcheese: can we assert: 5 toplevel declarations (or 3? idk if BinaryFile functions count)

bindata/bundle_test.go
wcharczuk: its quite a bit more complicated to generate this way, you achieve the same result with a simpler generator if you test that it compiles

bindata/bundle_test.go
wcharczuk: imports are much easier to sniff because they're treated specially vs. the file contents which can be a wild mess of ast nodes

bindata/byte_writer.go
```diff
@@ -0,0 +1,47 @@
...
+	c      int
+}
+
+func (w *ByteWriter) Write(p []byte) (n int, err error) {
+	if len(p) == 0 {
+		return
+	}
+
+	for n = range p {
+		if w.c%96 == 0 {
```
mat285: Not clear when reading why this number is here

PR #122, opened 2019-09-14
Selector parser option to extend permissible values.

nellamad:
## PR Summary

Implemented a parser option to extend the set of permissible chars/runes that can be matched against a value.  This is done by passing along a set of additionally permissible runes as optional parameters to the pre-existing validation methods.

We have a use case for this in smoke signal in which we'd like to match against labels that may contain an asterisk in its value.

 - **Type:** Features
 - **Issue Link:** https://blendlabs.atlassian.net/browse/INFRADEV-1862


wcharczuk:
i would not litter the predicate types with the maps as you have done, i would add an exported field to `Parser` that contains the rune array of possible values, and a corresponding option to set it. I would add a `RunsetOrDefault()` function that nil coalesces that field on the parser with the defaults, and then use that when checking the values.

nellamad:
I've updated the signature for `Selector.Validate()` to take an optional list of permitted values.  Adding a default runeset would require enumerating all of the runes that are currently validated against ranges of rune values. e.g. see `selector.isAlpha` and `selector.isSymbol`.  Should we still go that route?  I don't quite see the benefit.

selector/and.go
```diff
@@ -16,9 +16,9 @@ func (a And) Matches(labels Labels) bool {
 }
 
 // Validate validates all the selectors in the clause.
-func (a And) Validate() (err error) {
+func (a And) Validate(permittedValues ...map[rune]bool) (err error) {
```
wcharczuk: please make this an options object so we can just change that object if we need to add stuff later.

nellamad: good point

nellamad:
I reverted all of the selector changes.  The only thing left is the parser option to recognize `*` as a value instead of an operator.

selector/option.go
```diff
@@ -7,3 +7,16 @@ type Option func(p *Parser)
 func SkipValidation(p *Parser) {
 	p.skipValidation = true
 }
+
+// OptPermittedValues is an option to extend the set of symbols that are valid in name values
+func OptPermittedValues(permitted []rune) Option {
```
wcharczuk: Can we change this to be clearer about the intent; `OptExtraAlphas` and make it a variadic argument (i.e. `...rune`)

selector/util.go
```diff
@@ -48,6 +48,8 @@ const (
 	CarriageReturn = rune('\r')
 	// NewLine is a common rune.
 	NewLine = rune('\n')
+	// Asterisk is a common rune.
+	Asterisk = rune('*')
```
wcharczuk: Can you make this `Star`

selector/parser.go
```diff
@@ -27,7 +27,8 @@ type Parser struct {
 	// m is an optional mark
 	m int
 
-	skipValidation bool
+	skipValidation  bool
+	permittedValues map[rune]bool
```
wcharczuk: can we change this to `ExtraAlphas` and export `SkipValidation`, i.e. make it uppercase)

PR #123, opened 2019-09-18
request method variants with custom HTTP clients

afnanenayet:
## PR Summary

This extends some of the methods found in request so that users can supply their own HTTP clients which enables things like preventing requests from following redirects, and also allows you to expose all of the options from the go stdlib for HTTP requests.

 - **Type:** Improvements
 - **Issue Link:** n/a

#### Reviewers:


request/request.go
```diff
@@ -829,6 +883,26 @@ func (r *Request) BytesWithMeta() ([]byte, *ResponseMeta, error) {
 	return bytes, resMeta, nil
 }
 
+// BytesWithMetaClient fetches the response as bytes with meta, allowing the
+// user to supply a custom http client.
+func (r *Request) BytesWithMetaClient(client *http.Client) ([]byte, *ResponseMeta, error) {
```
nellamad: I would define a `WithClient` method to mutate the request object to stay consistent with this go-sdk version's style of adding options. (e.g. see `WithTracer` and `WithRequestHandler`).

afnanenayet: We could modify the `Request` struct so it holds an `http.Client`, which would be a more invasive change. The reason we can't currently just expose an option right now is because the client struct is hardcoded inside the function.

Because this is a change that would only be for v2 to support a tool that's going to be migrated to v3, I'm not sure how much effort is worth putting into this change, since modifying more things potentially breaks more things.

@wcharczuk thoughts?

wcharczuk: i'm with allen on, this, i would either define a `WithClient` fluent method or add a very tactical `WithRedirectHandler` method that sets the field on the client reference created when the response is fetched

PR #124, opened 2019-09-19
Selector Parser Options

wcharczuk:
## PR Summary

Adds many more knobs to turn with selector parsing.

 - **Type:** Features

#### Reviewers:


ex/exception.go
```diff
@@ -110,7 +110,7 @@ func (e *Ex) Format(s fmt.State, verb rune) {
 		if len(e.Message) > 0 {
 			io.WriteString(s, "; "+e.Message)
 		}
-		if s.Flag('+') {
+		if s.Flag('+') && e.StackTrace != nil {
```
dhermes: GitHub why you put this in the diff? It's already in `master`: https://github.com/blend/go-sdk/blob/6051c0b95d8d63733ac88802789a93a72b87ec07/ex/exception.go#L113

@wcharczuk Branch may need a rebase?

selector/check.go
```diff
@@ -0,0 +1,30 @@
...
+		if err = CheckValue(value); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// CheckKey validates a key with the default validation options.
+func CheckKey(key string) error {
+	return new(Parser).CheckKey(key)
```
dhermes: Any reason to prefer `new(Parser)` over `&Parser{}`? (Personally I avoid `new()`)

selector/check_test.go
```diff
@@ -0,0 +1,74 @@
+package selector
+
+import (
+	"strings"
+	"testing"
+
+	assert "github.com/blend/go-sdk/assert"
```
dhermes: Why the no-op import alias?

selector/selector.go
```diff
@@ -6,6 +6,5 @@ type Labels = map[string]string
 // Selector is the common interface for selector types.
 type Selector interface {
 	Matches(labels Labels) bool
-	Validate() error
```
dhermes: Why'd you nuke `Validate()`?

ex/exception.go
wcharczuk: this branch is super old

selector/check.go
wcharczuk: `new()` is clearer for me (specifically around if it will heap allocate)

selector/check_test.go
wcharczuk: `goimports` is annoying

selector/selector.go
wcharczuk: It's done when parsing, not on the final AST node

PR #127, opened 2019-09-24
Changes resulting from the work on blend/jobkit

wcharczuk:
## PR Summary

This PR overhauls the `jobkit` package and it's resulting `job` cli utility. It also makes a number of changes to the underlying `cron` package to enable some extra functionality like stats and history persistence.

 - **Type:** Features

#### Reviewers:


cron/job_invocation.go
```diff
@@ -2,31 +2,109 @@ package cron
...
+		State:          JobInvocationStateRunning,
+		JobName:        jobName,
+		Output:         output,
+		OutputHandlers: handlers,
+		Done:           make(chan struct{}),
 	}
 }
 
 // JobInvocation is metadata for a job invocation (or instance of a job running).
 type JobInvocation struct {
```
wcharczuk: one quick note: we put the output buffer and handlers here because they will eventually get serialized as history and it's nice to have the output as part of the history

bufferutil/buffer.go
```diff
@@ -0,0 +1,68 @@
...
+
+	// called outside critical section
+	if b.Handler != nil {
+		// call the handler with the chunk.
+		b.Handler(chunk)
+	}
+	return
+}
+
+// Bytes rerturns the bytes written to the writer.
```
mat285: `returns` typo

configutil/parse.go
```diff
@@ -23,6 +24,27 @@ type Parser struct {
...
+	value, err := p.Source.String()
+	if err != nil {
+		return nil, err
+	}
+	if value == nil {
+		return nil, nil
+	}
+	var parsed bool
+	switch strings.ToLower(*value) {
+	case "1", "true", "on", "yes":
```
mat285: should we add `enabled` ?

mat285: Also may want to list the valid values somewhere since its unclear without reading the code

cron/job_config_test.go
```diff
@@ -0,0 +1,33 @@
+package cron
+
+import (
+	"testing"
+	"time"
+
+	"github.com/blend/go-sdk/ref"
+
```
mat285: rm space

cron/job_interfaces.go
```diff
@@ -0,0 +1,133 @@
...
+/*
+A note on the naming conventions for the below interfaces.
+
+MethodName[Receiver|Provider] is the general pattern.
+
+"Receiver" indicates the function will be called by the manager.
+
+"Proivder" indicates the function will be called and is expected to return a specific value.
+
+They're mostly the same except
```
mat285: except what

cron/job_invocation.go
```diff
@@ -2,31 +2,109 @@ package cron
...
+	ji.ID = values.ID
+	ji.JobName = values.JobName
+	ji.Started = values.Started
+	ji.Finished = values.Finished
+	ji.Cancelled = values.Cancelled
+	ji.Timeout = values.Timeout
+	ji.Elapsed = values.Elapsed
+	ji.State = values.State
+	if values.Error != "" {
+		ji.Err = errors.New(values.Error)
```
mat285: why errors instead of ex

cron/job_manager.go
```diff
@@ -115,20 +125,20 @@ func (jm *JobManager) IsJobDisabled(jobName string) (value bool) {
...
 
-// IsJobRunning returns if a task is currently running.
+// IsJobRunning returns if a job is currently running.
 func (jm *JobManager) IsJobRunning(jobName string) (isRunning bool) {
 	jm.Lock()
 	defer jm.Unlock()
 
 	if job, ok := jm.Jobs[jobName]; ok {
-		isRunning = job.Current != nil
+		isRunning = !job.Idle()
```
mat285: This seems to be a different idea of "running". Seems like it used to be is the job "currently executing" vs "not idle"

cron/job_scheduler_option.go
```diff
@@ -14,8 +14,3 @@ func OptJobSchedulerTracer(tracer Tracer) JobSchedulerOption {
 func OptJobSchedulerLog(log logger.Log) JobSchedulerOption {
 	return func(js *JobScheduler) { js.Log = log }
 }
-
-// OptJobSchedulerConfig sets the job scheduler config.
-func OptJobSchedulerConfig(hc Config) JobSchedulerOption {
```
mat285: Meant to delete?

examples/reverseproxy/main.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/blend/go-sdk/logger"
+	"github.com/blend/go-sdk/r2"
+	"github.com/blend/go-sdk/reverseproxy"
+	"github.com/blend/go-sdk/webutil"
+)
+
+func main() {
+	log := logger.Prod()
+
+	_, err := r2.New("https://sentry.k8s.tools.blend.com").CopyTo(os.Stdout)
```
mat285: this fine being public?

configutil/parse.go
wcharczuk: moved to a central function in `stringutil`

cron/job_invocation.go
wcharczuk: when we're unmarshalling we just want to trivially wrap this as an error, i'm not sure we want to do a full ex.Unmarshal

cron/job_scheduler_option.go
wcharczuk: yeah job schedulers don't have configs

examples/reverseproxy/main.go
wcharczuk: that was from a different pr, agreed it shouldn't be public

cron/job_manager.go
wcharczuk: they're effectively the same thing, it's just the name is clearer

env/util.go
```diff
@@ -13,7 +13,7 @@ import (
 func mustParseBool(str string) bool {
 	strLower := strings.ToLower(str)
 	switch strLower {
-	case "true", "1", "yes":
+	case "true", "1", "yes", "enabled":
```
mat285: Nit: Can you use `parseBool` to implement `mustParseBool` instead of duplicating conditionals?

env/util.go
```diff
@@ -22,9 +22,9 @@ func mustParseBool(str string) bool {
 func parseBool(str string) (bool, error) {
 	strLower := strings.ToLower(str)
 	switch strLower {
-	case "true", "1", "yes":
+	case "true", "1", "yes", "enabled":
```
mat285: actually second q, why is this not using stringutil?

wcharczuk: because i didn't save the file, good catch

cron/job_scheduler.go
```diff
@@ -550,7 +558,9 @@ func (js *JobScheduler) Enabled() bool {
 
 // Idle returns if the job is not currently running.
 func (js *JobScheduler) Idle() (isIdle bool) {
+	js.Lock()
 	isIdle = js.Current == nil
+	js.Unlock()
```
mat285: v small thing, but defer js.Unlock() for consistency?

cron/job_scheduler_stats.go
```diff
@@ -0,0 +1,19 @@
...
+type JobSchedulerStats struct {
+	JobName        string        `json:"jobName"`
+	SuccessRate    float64       `json:"successRate"`
+	OutputBytes    int           `json:"outputBytes"`
+	RunsTotal      int           `json:"runsTotal"`
+	RunsSuccessful int           `json:"runsSuccessful"`
+	RunsFailed     int           `json:"runsFailed"`
+	RunsCancelled  int           `json:"runsCancelled"`
+	RunsTimedOut   int           `json:"runsTimedOut"`
+	ElapsedMax     time.Duration `json:"elapsedMax"`
```
mikes-blend: Why are we tracking all this in code instead of just plugging execution time into a histogram and letting DD figure it out?

mat285: I'm assuming for the jobkit ui and for those that don't use dd

wcharczuk: yeah it's because this is meant to be self contained and these are pretty easy to calculate

cron/job_scheduler.go
wcharczuk: saves a stack frame for the defer

PR #132, opened 2019-09-30
Removes webutil/request_sender, switches slack to use r2

wcharczuk:
## PR Summary

* Removes webutil/request_sender, switches slack to use r2.
* Adds event source helper for writing event source output

 - **Type:** Internal

#### Reviewers:


slack/webhook_sender.go
```diff
@@ -5,10 +5,11 @@ import (
 	"encoding/json"
 	"io/ioutil"
 
+	"github.com/blend/go-sdk/r2"
```
mat285: move this into the right import block

PR #134, opened 2019-10-04
Coverage CLI Tweaks

wcharczuk:
## PR Summary

Adds ability to specify path roots in `./...` format. Switches include and exclude statements to be arrays (i.e. you can specify more than one). Also allows the `-race` flag now.

One caveat, you must invoke it with `./...` args last, i.e.:

```bash
> coverage -race -timeout=10s --include="*.go" --exclude="_testdata/*" ./yaml/... ./stats/...
```

Should be strictly backwards compatible.

 - **Type:** Improvements

#### Reviewers:


cmd/coverage/main.go
```diff
@@ -34,31 +34,114 @@ const (
...
 var coverprofile = flag.String("coverprofile", "coverage.cov", "the intermediate cover profile.")
 var keepCoverageOut = flag.Bool("keep-coverage-out", false, "if we should keep coverage.out")
 var v = flag.Bool("v", false, "show verbose output")
 
-func verbose() bool {
-	if v != nil && *v {
-		return true
+var (
+	includes paths
+	excludes paths
```
dhermes: Sidebar: There is a local in `main()` called `paths` and a type called `paths`

cmd/coverage/main.go
```diff
@@ -89,17 +172,21 @@ func getPackageCoverage(currentPath string, info os.FileInfo, err error) (string
 		return "", nil
 	}
 
-	if len(*include) > 0 {
-		if matches := globAnyMatch(*include, currentPath); !matches {
-			vf("`%s` skipping dir; include no match: %s", currentPath, *include)
-			return "", nil
+	if len(includes) > 0 {
```
dhermes: The length check is superfluous, if empty, the range will no-op (similar for line 184)

cmd/coverage/main.go
```diff
@@ -89,17 +172,21 @@ func getPackageCoverage(currentPath string, info os.FileInfo, err error) (string
...
+		for _, include := range includes {
+			if matches := globAnyMatch(include, currentPath); !matches { // note the !
+				vf("`%s` skipping dir; include no match: %s", currentPath, include)
```
dhermes: Use `%q` instead of backticks around `%s`?

cmd/coverage/main_test.go
```diff
@@ -28,13 +29,13 @@ func TestGlob(t *testing.T) {
...
 }
+
 func TestGlobAnyMatch(t *testing.T) {
 	assert := assert.New(t)
 
 	assert.False(globAnyMatch("", "testo"))
 	assert.True(globAnyMatch("*", "testo"))
-	assert.True(globAnyMatch("x,testo", "testo"))
-	assert.False(globAnyMatch("x,y", "testo"))
+	assert.False(globAnyMatch("x", "testo"))
```
dhermes: `globAnyMatch()` hasn't changed, but our need for it to account for commas **has** changed. I think we can safely delete it.

cmd/coverage/main.go
```diff
@@ -89,17 +172,21 @@ func getPackageCoverage(currentPath string, info os.FileInfo, err error) (string
 		return "", nil
 	}
 
-	if len(*include) > 0 {
-		if matches := globAnyMatch(*include, currentPath); !matches {
-			vf("`%s` skipping dir; include no match: %s", currentPath, *include)
-			return "", nil
+	if len(includes) > 0 {
+		for _, include := range includes {
+			if matches := globAnyMatch(include, currentPath); !matches { // note the !
```
dhermes: This should just be `glob(inlcude, currentPath)`? Ditto on 186

cmd/coverage/main.go
```diff
@@ -29,36 +29,120 @@ var (
...
-}
+	fmt.Fprintf(fullCoverageData, "mode: %s\n", *covermode)
 
-func verrf(format string, args ...interface{}) {
-	if verbose() {
-		fmt.Fprintf(os.Stderr, "coverage :: err :: "+format+"\n", args...)
+	paths := flag.Args()
+
+	if len(paths) == 0 {
+		paths = []string{"./..."}
```
dhermes: `"." + expand`?

dhermes:
Ping me when `cron` COVERAGE collision is handled?

PR #136, opened 2019-11-06
Adds etags for web static cached results

wcharczuk:
## PR Summary

Adds etag functionality to the static file server in cached mode.

 - **Type:** Improvements

#### Reviewers:


web/static_file_server.go
```diff
@@ -127,6 +130,13 @@ func (sc *StaticFileServer) ServeFile(r *Ctx, filePath string) Result {
 
 	finfo, err := f.Stat()
 	if err != nil {
+		if os.IsNotExist(err) {
```
dhermes: 1. This change seems unrelated to the ETag change?
1. These lines are identical to 155-167, thoughts on factoring out?

webutil/etag.go
```diff
@@ -0,0 +1,18 @@
...
+
+	"github.com/blend/go-sdk/ex"
+)
+
+// ETag creates an etag for a given blob.
+func ETag(contents []byte) (string, error) {
+	hash := md5.New()
+	_, err := hash.Write(contents)
+	if err != nil {
+		return "", ex.New(err)
```
dhermes: Seems quite challenging to cook up a `contents` that will trigger this branch?

mat285: This actually never will return an error according to the docs so feel free to not check it https://golang.org/pkg/hash/

mat285: It only has it in the spec because of `io.Writer`

webutil/etag.go
```diff
@@ -3,16 +3,11 @@ package webutil
...
 // ETag creates an etag for a given blob.
-func ETag(contents []byte) (string, error) {
+func ETag(contents []byte) string {
 	hash := md5.New()
-	_, err := hash.Write(contents)
-	if err != nil {
-		return "", ex.New(err)
-	}
-	return hex.EncodeToString(hash.Sum(nil)), nil
+	hash.Write(contents)
```
dhermes: ISTM `_, _ = hash.Write(contents)` is the better move here, with a comment explaining why it's safe to ignore the error.

If we ever decide to integrate [`errcheck`][1] into CI (which I highly recommend) this will get flagged.

[1]: https://github.com/kisielk/errcheck

Issue #142, opened 2019-11-15
Any r2/webutil option that modifies the body should also set content length


dhermes:
There is a [large effort][1] in `http.NewRequestWithContext` to set `req.ContentLength` and `req.GetBody` whenever possible (and `req.Body` [is added][3] in the `Request{}` struct literal). However, our `webutil` options [just set][2] `req.Body` directly.

Related to #141.

[1]: https://github.com/golang/go/blob/go1.13.1/src/net/http/request.go#L872-L914
[2]: https://github.com/blend/go-sdk/blob/7b1966b8f401f28a23a8ba565e67a1d241643547/webutil/request_option.go#L239
[3]: https://github.com/golang/go/blob/go1.13.1/src/net/http/request.go#L869

PR #148, opened 2019-12-03
Setting content length in many cases where `req.Body` is set.

dhermes:
## PR Summary

This does not necessarily cover **all** cases where we set `req.Body`, but it covers most. This partially addresses #142, where we are sending chunked HTTP requests unwittingly.

 - **Type:** Bugfix
 - **Issue Link:** https://github.com/blend/go-sdk/issues/142

### DIFF
--- a/r2/request.go
+++ b/r2/request.go
@@ -63,12 +63,10 @@ func (r Request) Do() (*http.Response, error) {
 	}
 
 	// reconcile post form values
-	if r.Request.PostForm != nil && len(r.Request.PostForm) > 0 {
-		if r.Request.Body == nil {
-			body := r.Request.PostForm.Encode()
-			r.Request.Body = ioutil.NopCloser(strings.NewReader(body))
-			r.Request.ContentLength = int64(len(body))
-		}
+	if r.Request.PostForm != nil && len(r.Request.PostForm) > 0 && r.Request.Body == nil {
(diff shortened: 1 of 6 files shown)


r2/request.go
```diff
@@ -65,7 +65,9 @@ func (r Request) Do() (*http.Response, error) {
 	// reconcile post form values
 	if r.Request.PostForm != nil && len(r.Request.PostForm) > 0 {
 		if r.Request.Body == nil {
```
wcharczuk: we should probably move this to be && above

webutil/request_option.go
```diff
@@ -193,6 +193,7 @@ func OptBody(contents io.ReadCloser) RequestOption {
 func OptBodyBytes(body []byte) RequestOption {
 	return func(r *http.Request) error {
 		r.Body = ioutil.NopCloser(bytes.NewBuffer(body))
+		r.ContentLength = int64(len(body))
```
wcharczuk: do we need to set the header? is the content length field on the outgoing request sufficient?

dhermes: Based on https://github.com/golang/go/blob/go1.13.1/src/net/http/request.go#L872-L914, I'd say no

wcharczuk: got it, lgtm

PR #155, opened 2019-12-14
Fix job history leak

chotiwat:
## PR Summary

The job history is enabled by default and the default max age and max count aren't set. This results in a default behavior of keeping the history of job invocations forever which results in a memory leak. This PR disables job history by default and provides default max age and max count when they're unset.

 - **Type:** Bugfix

#### Reviewers:

### DIFF
--- a/cron/job_scheduler.go
+++ b/cron/job_scheduler.go
@@ -588,7 +588,7 @@ func (js *JobScheduler) finishCurrent(ji *JobInvocation) {
 	js.Lock()
 	defer js.Unlock()
 
-	if !js.HistoryEnabledProvider() {
+	if js.HistoryEnabledProvider() {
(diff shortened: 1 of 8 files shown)


cron/job_scheduler.go
```diff
@@ -588,7 +588,7 @@ func (js *JobScheduler) finishCurrent(ji *JobInvocation) {
 	js.Lock()
 	defer js.Unlock()
 
-	if !js.HistoryDisabledProvider() {
+	if !js.HistoryEnabledProvider() {
```
wcharczuk: i think this is wrong; it should be `if js.HistoryEnabledProvider()`

chotiwat: Whoops, yep, I find&replace them all 😳

PR #157, opened 2019-12-16
Making sure `web.Ctx.CookieDomain()` does not include port.

### LINKED ISSUE #156: v3: `web.Ctx.CookieDomain()` includes port
This leads to

```
...
2019-12-16 11:17:14.441421 I | net/http: invalid Cookie.Domain "localhost:8080"; dropping domain attribute
...
```

See [an example][1], we are [using][2] `u.Host` rather than `u.Hostname()`:

```
=========================================
u.Host:       "localhost:8080"
u.Hostname(): "localhost"
Raw URL:      "http://localhost:8080"
Parsed URL:   &url.URL{Scheme:"http", Opaque:"", User:(*url.Userinfo)(nil), Host:"localhost:8080", Path:"", RawPath:"", ForceQuery:false, RawQuery:"", Fragment:""}
=========================================
u.Host:       "web.invalid"
u.Hostname(): "web.invalid"
Raw URL:      "https://web.invalid"
Parsed URL:   &url.URL{Scheme:"https", Opaque:"", User:(*url.Userinfo)(nil), Host:"web.invalid", Path:"", RawPath:"", ForceQuery:false, RawQuery:"", Fragment:""}
```

(It's going to take me less time to write a fix than it did filing this.)

[1]: https://play.golang.org/p/kIgWUP_k1qu
[2]: https://github.com/blend/go-sdk/blob/480ad70d161406204ef58d3528b0ac17cf64bf0a/web/ctx.go#L262-L267

dhermes:
## PR Summary

Fixes #156

 - **Type:** Bugfix
 - **Issue Link:** https://github.com/blend/go-sdk/issues/156

### DIFF
--- a/web/ctx.go
+++ b/web/ctx.go
@@ -265,9 +265,7 @@ func (rc *Ctx) CookieDomain() string {
 		u := webutil.MustParseURL(rc.App.Config.BaseURL)
 		return u.Hostname()
 	}
-	// Strip off the port if the hostname includes it.
-	parts := strings.Split(rc.Request.Host, ":")
-	return parts[0]
+	return extractHost(rc.Request.Host)
 }
 
 // Cookie returns a named cookie from the request.
@@ -396,4 +394,22 @@ func (rc *Ctx) loggerAnnotations() logger.Annotations {
 		fields["web.route_parameters"] = rc.RouteParams
 	}
 	return logger.CombineAnnotations(logger.GetAnnotations(rc.Request.Context()), fields)
+}
+
+// extractHost splits a host / port pair (or just a host) and returns the host.
+// This is large borrowed from `net/url.splitHostPort` (as of `go1.13.5`).
+func extractHost(hostport string) string {
+	host := hostport
+
+	colon := strings.LastIndexByte(host, ':')
+	if colon != -1 {
+		host = host[:colon]
+	}
+
+	// If `hostport` is an IPv6 address of the form `[::1]:12801`.
+	if strings.HasPrefix(host, "[") && strings.HasSuffix(host, "]") {
+		host = host[1 : len(host)-1]
+	}
+
+	return host
 }
--- a/web/ctx.go
+++ b/web/ctx.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/blend/go-sdk/ex"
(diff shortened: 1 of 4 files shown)


web/ctx.go
```diff
@@ -261,9 +262,12 @@ func (rc *Ctx) PostBodyAsXML(response interface{}) error {
 // CookieDomain returns the cookie domain for a request.
 func (rc *Ctx) CookieDomain() string {
 	if rc.App != nil && rc.App.Config.BaseURL != "" {
-		return webutil.MustParseURL(rc.App.Config.BaseURL).Host
+		u := webutil.MustParseURL(rc.App.Config.BaseURL)
```
wcharczuk: why not just `s/.Host/.Hostname()`?

web/ctx.go
```diff
@@ -261,9 +262,12 @@ func (rc *Ctx) PostBodyAsXML(response interface{}) error {
...
+		return u.Hostname()
 	}
-	return rc.Request.Host
+	// Strip off the port if the hostname includes it.
+	parts := strings.Split(rc.Request.Host, ":")
```
dhermes: Thoughts on just doing what `net/url` does [in `splitHostPort`][1]?

[1]: https://github.com/golang/go/blob/go1.13.5/src/net/url/url.go#L1071

web/ctx.go
dhermes: See my comment on https://github.com/blend/go-sdk/issues/156#issue-538615678. TL;DR is [they are different][1]

```
=========================================
u.Host:       "localhost:8080"
u.Hostname(): "localhost"
Raw URL:      "http://localhost:8080"
Parsed URL:   &url.URL{Scheme:"http", Opaque:"", User:(*url.Userinfo)(nil), Host:"localhost:8080", Path:"", RawPath:"", ForceQuery:false, RawQuery:"", Fragment:""}
=========================================
u.Host:       "web.invalid"
u.Hostname(): "web.invalid"
Raw URL:      "https://web.invalid"
Parsed URL:   &url.URL{Scheme:"https", Opaque:"", User:(*url.Userinfo)(nil), Host:"web.invalid", Path:"", RawPath:"", ForceQuery:false, RawQuery:"", Fragment:""}
```

[1]: https://play.golang.org/p/kIgWUP_k1qu

web/ctx.go
dhermes: We could make it even smaller though because we wouldn't need to check if `:port` is valid?

dhermes: Ooooh, TIL `net.SplitHostPort` exists ([example][1]) but it **requires** a port:

```
=================================
hostport: "localhost:8080"
host:     "localhost"
port:     "8080"
err:      <nil>
=================================
hostport: "web.invalid"
host:     ""
port:     ""
err:      &net.AddrError{Err:"missing port in address", Addr:"web.invalid"}
=================================
hostport: "[::1]:16789"
host:     "::1"
port:     "16789"
err:      <nil>
```

[1]: https://play.golang.org/p/kA0eZlp8uJT

PR #158, opened 2019-12-19
Do not log broken pipe errors when rendering results

nellamad:
## PR Summary

The web package is logging broken pipe errors as fatal errors which triggers an issue in sentry.  This PR skips logging for broken pipe errors returned after rendering a Result.  Tested this manually by running Deployinator locally and inducing broken pipe errors with and without this added check.

 - **Type:** Improvements
 - **Issue Link:** 

#### Reviewers:

### DIFF
--- a/web/app.go
+++ b/web/app.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"net/http"
 	"strings"
+	"syscall"
 
 	"github.com/blend/go-sdk/async"
 	"github.com/blend/go-sdk/ex"
@@ -414,8 +415,8 @@ func (a *App) RenderAction(action Action) Handler {
 				}
 			}
 
-			// do the render, log any errors emitted
-			if resultErr := result.Render(ctx); resultErr != nil {
+			// do the render, log any errors emitted except broken pipe errors
+			if resultErr := result.Render(ctx); resultErr != nil && !ex.Is(resultErr, syscall.EPIPE) {
 				err = ex.Nest(err, resultErr)
 				a.maybeLogFatal(ctx.Context(), resultErr, ctx.Request)
 			}
--- a/web/app_test.go
+++ b/web/app_test.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"strings"
 	"sync"
+	"syscall"
 	"testing"
 	"time"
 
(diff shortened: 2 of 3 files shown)


web/app.go
```diff
@@ -414,8 +415,8 @@ func (a *App) RenderAction(action Action) Handler {
 				}
 			}
 
-			// do the render, log any errors emitted
-			if resultErr := result.Render(ctx); resultErr != nil {
+			// do the render, log any errors emitted except broken pipe errors
+			if resultErr := result.Render(ctx); resultErr != nil && !ex.Is(resultErr, syscall.EPIPE) {
```
nellamad: side note: Go's `errors` package seems to support the [`Is` predicate](https://golang.org/pkg/errors/#Is) as of version 1.13 so we may want to think about updating our `ex.Is`

dhermes: `syscall.EPIPE` is **probably** not portable. @wcharczuk what's our stance on this?

web/app_test.go
```diff
@@ -440,6 +449,33 @@ func TestAppNotFound(t *testing.T) {
...
+
+	app, err := New(OptLog(agent))
+	assert.Nil(err)
+
+	wg := sync.WaitGroup{}
+	wg.Add(1)
+
+	app.GET("/", func(r *Ctx) Result {
+		defer wg.Done()
+		result := BrokenPipeResult("I'm a broken pipe result!")
```
dhermes: Here is my handle, here is my spout

web/app_test.go
```diff
@@ -440,6 +449,33 @@ func TestAppNotFound(t *testing.T) {
...
+		return &result
+	})
+
+	_, err = MockGet(app, "/").Discard()
+	assert.Nil(err)
+	wg.Wait()
+
+	// Test: verify that the broken pipe error was not logged
+	assert.Nil(agent.Drain())
+	assert.Equal(-1, strings.Index(buffer.String(), ex.New(syscall.EPIPE).Error()))
```
dhermes: Why can't you just do an assert on the value of `buffer.String()`, isn't it `""`?

nellamad:
Closing this since the root issue is better solved by controlling what events Deployinator forwards to Sentry.

PR #159, opened 2020-01-08
Fix LRU Ordering & Sweep after Local Cache Updates

akarimcheese:
## PR Summary

Holds off on updating value on reference shared by Cache's underlying map and LRU queue until after the LRU has processed the update.

 - **Type:** Bugfix

#### Reviewers:

### DIFF
--- a/cache/local_cache_test.go
+++ b/cache/local_cache_test.go
@@ -166,6 +166,27 @@ func TestLocalCacheGetOrSetDoubleCheckRace(t *testing.T) {
 	assert.Equal("bar2", found)
 }
 
+func TestLocalCacheSetUpdatesLRU(t *testing.T) {
+	assert := assert.New(t)
+
+	c := NewLocalCache()
+	c.Set("k1", "v1", OptValueTTL(0))
+	time.Sleep(time.Millisecond)
+	c.Set("k2", "v2", OptValueTTL(0))
+	assert.Equal("k1", c.LRU.Peek().Key)
+
+	// Should trigger sorting of underlying LRU so k2 can be
+	// deleted in next sweep
+	time.Sleep(time.Millisecond)
+	c.Set("k1", "v3", OptValueTTL(time.Second))
+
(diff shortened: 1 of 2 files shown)


akarimcheese:
Old code resulted in this always evaluating the true

https://github.com/blend/go-sdk/blob/8de5c48bd4f72dca0e387f9477c42e56f89132db/cache/lru_queue.go#L107

and thus never triggering a sort of underlying LRU

cache/local_cache_test.go
```diff
@@ -166,6 +166,26 @@ func TestLocalCacheGetOrSetDoubleCheckRace(t *testing.T) {
 	assert.Equal("bar2", found)
 }
 
+func TestLocalCacheSetUpdatesLRU(t *testing.T) {
+	assert := assert.New(t)
+
+	c := NewLocalCache()
+	c.Set("k1", "v1", OptValueTTL(0))
+	time.Sleep(time.Millisecond)
```
dhermes: Why this sleep?

cache/local_cache_test.go
```diff
@@ -166,6 +166,26 @@ func TestLocalCacheGetOrSetDoubleCheckRace(t *testing.T) {
...
+	c.Set("k2", "v2", OptValueTTL(0))
+	assert.Equal("k1", c.LRU.Peek().Key)
+
+	// Should trigger sorting of underlying LRU so k2 can be
```
dhermes: "Should trigger" refers to `time.Sleep()` or `c.Set()`?

akarimcheese: Trigger refers to c.Set(). I'll move the sleep above the comment

cache/local_cache_test.go
akarimcheese: Didn't realize the initial `Set()` for a key was an LRU `Push()` instead of a `Fix()`, so this isn't necessary. Will remove

PR #162, opened 2020-01-15
Adds Sentry `Fingerprint` support

wcharczuk:
Adds fingerprinting support to the Sentry SDK bindings

### DIFF
--- a/sentry/client.go
+++ b/sentry/client.go
@@ -86,6 +86,9 @@ func errEvent(ctx context.Context, ee logger.ErrorEvent) *raven.Event {
 }
 
 func errFingerprint(ctx context.Context, extra ...string) []string {
+	if fingerprint := GetFingerprint(ctx); fingerprint != nil {
+		return fingerprint
+	}
 	return append(logger.GetPath(ctx), extra...)
 }
 
--- a/sentry/client.go
+++ b/sentry/client.go
@@ -59,11 +59,12 @@ func (c Client) Notify(ctx context.Context, ee logger.ErrorEvent) {
 
 func errEvent(ctx context.Context, ee logger.ErrorEvent) *raven.Event {
 	return &raven.Event{
-		Timestamp: logger.GetEventTimestamp(ctx, ee).Unix(),
-		Level:     raven.Level(ee.GetFlag()),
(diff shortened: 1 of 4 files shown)


sentry/client.go
```diff
@@ -84,6 +85,13 @@ func errEvent(ctx context.Context, ee logger.ErrorEvent) *raven.Event {
 	}
 }
 
+func errFingerprint(ctx context.Context, extra ...string) []string {
+	if fingerprint := GetFingerprint(ctx); fingerprint != nil {
+		return fingerprint
```
dhermes: OK that `GetFingerprint()` doesn't use the `extra` string slice?

sentry/client_test.go
```diff
@@ -56,6 +56,7 @@ func TestErrEvent(t *testing.T) {
 	assert.Equal(SDK, event.Sdk.Name)
 	assert.Equal("this is a test", event.Message)
 	assert.NotEmpty(event.Exception)
+	assert.NotEmpty(event.Fingerprint)
```
dhermes: Why not check the value?

sentry/client_test.go
```diff
@@ -84,3 +85,63 @@ func TestErrFrames(t *testing.T) {
...
+			Method: "POST",
+			Host:   "example.org",
+			TLS:    &tls.ConnectionState{},
+			URL:    webutil.MustParseURL("https://example.org/foo"),
+		},
+	})
+
+	assert.NotNil(event)
+	assert.NotEmpty(event.Fingerprint)
+	assert.Equal("this is a test", event.Fingerprint[0])
```
dhermes: Nit: I'd prefer

```suggestion
	assert.Equal([]string{"this is a test"}, event.Fingerprint)
```

as future-proofing against that case the `event.Fingerprint` has more than 1 element.

dhermes:
FYI @wcharczuk I rebased this after #161 merged (trying to clean up my own disruption). If your local branch won't push because of "fast forward" issues, that's why.

sentry/client_test.go
wcharczuk: there is more detailed test for this later

PR #164, opened 2020-01-15
Adding `stats/tracing/httptrace` package.

dhermes:
## PR Summary

 - **Type:** Features

#### Reviewers:

This is a lower-level trace interface than `webtrace`, but they share most of the same ideas, so `webtrace.webTracer.Start()` shares almost all of it's code with `httptrace.httpTracer.Start()` via `httptrace.StartHTTPSpan()`


web/tracer.go
```diff
@@ -19,3 +23,15 @@ type ViewTracer interface {
...
+// `(net/http).Request`. It returns a "new" request the request context may
+// be modified after opening a span.
+type HTTPTracer interface {
+	Start(*http.Request) (HTTPTraceFinisher, *http.Request)
+}
+
+// HTTPTraceFinisher is a simplified version of `TraceFinisher`.
+type HTTPTraceFinisher interface {
+	Finish(error)
+}
```
dhermes: I'm open to putting `web.HTTPTracer` and `web.HTTPTraceFinisher` if people have thoughts.

stats/tracing/httptrace/tracer.go
```diff
@@ -0,0 +1,78 @@
...
+
+	opentracing "github.com/opentracing/opentracing-go"
+
+	"github.com/blend/go-sdk/stats/tracing"
+	"github.com/blend/go-sdk/web"
+	"github.com/blend/go-sdk/webutil"
+)
+
+var (
+	_ web.HTTPTracer        = (*httpTracer)(nil)
```
wcharczuk: i would not have a reference to `web` if possible here, you can define the interface in something more low level like `webutil` if you do want it in a shared place

web/tracer.go
wcharczuk: i would move these to `webutil`

dhermes:
Done. @wcharczuk PTAL

stats/tracing/httptrace/tracer.go
```diff
@@ -0,0 +1,77 @@
...
+	tracer opentracing.Tracer
+}
+
+// StartHTTPSpan opens a span and creates a new request with a modified
+// context, based on the span that was opened.
+func StartHTTPSpan(ctx context.Context, tracer opentracing.Tracer, req *http.Request, resource string, startTime time.Time, extra ...opentracing.StartSpanOption) (opentracing.Span, *http.Request) {
+	// set up basic start options (these are mostly tags).
+	startOptions := []opentracing.StartSpanOption{
+		opentracing.Tag{Key: tracing.TagKeyResourceName, Value: resource},
+		opentracing.Tag{Key: tracing.TagKeySpanType, Value: tracing.SpanTypeWeb},
```
dhermes: Just realized I have this copied over. `"web"` probably doesn't make sense for pure HTTP

stats/tracing/httptrace/tracer.go
```diff
@@ -0,0 +1,77 @@
...
+
+	// try to extract an incoming span context
+	// this is typically done if we're a service being called in a chain from another (more ancestral)
+	// span context.
+	spanContext, _ := tracer.Extract(opentracing.TextMap, opentracing.HTTPHeadersCarrier(req.Header))
+	if spanContext != nil {
+		startOptions = append(startOptions, opentracing.ChildOf(spanContext))
+	}
+	// start the span.
+	span, spanCtx := tracing.StartSpanFromContext(ctx, tracer, tracing.OperationHTTPRequest, startOptions...)
```
dhermes: Similar to the above comment, having the operation name **always** be `http.request` may be something we'd allow to tweak. I'm making this change so I can use this for `reverseproxy` (#165), so making operation name something else, e.g. `http.proxied_request`, may be useful.

stats/tracing/httptrace/tracer.go
wcharczuk: spanTypeWeb is a thing in the dd nomenclature, it roughly translates to http

stats/tracing/httptrace/tracer.go
wcharczuk: i'd want to dig into exactly what the impact of a name change like that would be but imho it's fine to use whatever here as long as we can make heads or tails of it later

stats/tracing/httptrace/tracer.go
dhermes: Thanks!

stats/tracing/httptrace/tracer.go
dhermes: SG, will cross that bridge when we get there.

PR #168, opened 2020-01-17
Adding opaque `TransformRequest` to `reverseproxy.Proxy`

dhermes:
## PR Summary

The current set of supported options only allow direct interaction with headers, but this feature allows for more generic actions based on the request. For example, reading `req.TLS.PeerCertificates` to get information about the client certificate used in mTLS.

 - **Type:** Features


reverseproxy/transform_request.go
```diff
@@ -0,0 +1,10 @@
+package reverseproxy
+
+import (
+	"net/http"
+)
+
+// TransformRequest modifies an HTTP request. This is intended to be used
+// during `Proxy.ServeHTTP()` for custom business logic, e.g. checking if a
+// client was included and verified in the request.
+type TransformRequest = func(*http.Request) error
```
dhermes: It's kind of silly to introduce this file just for a type. Any good suggestions? Other files in this package containing `type`-s:

```
$ git grep -n 'type ' -- '*.go' ':(exclude)*_test.go'
proxy.go:29:type Proxy struct {
proxy_option.go:10:type ProxyOption func(*Proxy) error
redirect_http.go:10:type HTTPRedirect struct{}
resolver.go:9:type Resolver func(*http.Request, []*Upstream) (*Upstream, error)
transform_request.go:10:type TransformRequest = func(*http.Request) error
upstream.go:25:type Upstream struct {
```

reverseproxy/proxy.go
```diff
@@ -79,6 +80,12 @@ func (p *Proxy) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 			req.Header.Add(key, value)
 		}
 	}
+	if p.TransformRequest != nil {
+		err := p.TransformRequest(req)
```
dhermes: If we wanted to make this **more** locked down, we could pass a **value** `*req` and `req.Header` so that way `TransformRequest` could **READ** the full contents of `req` but could only **WRITE** to `req.Header` (which is what we want an L7 reverse proxy to do).

reverseproxy/proxy.go
```diff
@@ -79,6 +80,12 @@ func (p *Proxy) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+		if err != nil {
+			logger.MaybeError(p.Log, err)
```
dhermes: Is it worth exiting here as well? Other places where an error occurs return a 502, e.g.

```go
	if err != nil {
		logger.MaybeError(p.Log, err)
		rw.WriteHeader(http.StatusBadGateway)
		return
	}
```

reverseproxy/proxy.go
dhermes: `(net/http/httputil).ReverseProxy.Director`would be the [equivalent][1]

[1]: https://github.com/golang/go/blob/go1.13.6/src/net/http/httputil/reverseproxy.go#L28-L34 would

reverseproxy/transform_request.go
mat285: I'd just keep it a separate file

reverseproxy/proxy.go
```diff
@@ -105,6 +106,12 @@ func (p *Proxy) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 			req.Header.Add(key, value)
 		}
 	}
+	if p.TransformRequest != nil {
+		err = p.TransformRequest(req)
```
dhermes: If we wanted to make this **more** locked down, we could pass a **value** `*req` and `req.Header` so that way `TransformRequest` could **READ** the full contents of `req` but could only **WRITE** to `req.Header` (which is what we want an L7 reverse proxy to do).

`(net/http/httputil).ReverseProxy.Director` would be the [equivalent][1]

[1]: https://github.com/golang/go/blob/go1.13.6/src/net/http/httputil/reverseproxy.go#L28-L34

reverseproxy/proxy.go
dvdliao: adjusting the func prototype to not return makes most sense as we're not doing anything with the error here

PR #165, opened 2020-01-15
Adding optional `Tracer` to `reverseproxy.Proxy`.

dhermes:
## PR Summary

 - **Type:** Features
 - **Intended Change Level:** minor

**NOTE**: This has #164 as diffbase (though since it's in my fork, I can't make it the diffbase of this PR). As of 3:15pm on January 15, 2020 [this][1] view will give the 4 commits relevant to this PR.

[1]: https://github.com/blend/go-sdk/pull/165/files/95f2209b3917ed6ca2dae6b5196e8a2e5230926d..02c1875cfed72deb264691d4a6c3f2dae22f8ae7

PR #169, opened 2020-01-17
Setting `ErrorHandler` on `(net/http/httputil).ReverseProxy`.

dhermes:
## PR Summary

Without it the default error handler is used, which ends up logging
directly to STDOUT, e.g. for a failed DNS lookup

```
2020/01/17 13:31:06 http: proxy error: dial tcp: lookup web.invalid: no such host
```

and for a bad client certificate presented as part of a TLS handshake:

```
2020/01/17 00:52:35 http: TLS handshake error from 172.30.0.1:51932: tls: failed to verify client's certificate: x509: certificate signed by unknown authority
```

I noticed this initially during unit tests in #165 and #168.

 - **Type:** Improvements


reverseproxy/upstream.go
```diff
@@ -15,10 +15,14 @@ import (
...
 // NewUpstream returns a new upstram.
 func NewUpstream(target *url.URL) *Upstream {
-	return &Upstream{
+	rp := httputil.NewSingleHostReverseProxy(target)
+	u := &Upstream{
 		URL:          target,
-		ReverseProxy: httputil.NewSingleHostReverseProxy(target),
+		ReverseProxy: rp,
 	}
+	// NOTE: This creates a reference cycle `u -> rp -> u`.
```
dhermes: I toyed with making a closure that just references `u.Log`, but at this time `u.Log` isn't set yet, so that is a no-go.

mat285:
What's the benefit of taking this route rather than having user's write it themselves? Or providing it as an option instead?

dhermes:
> What's the benefit of taking this route rather than having user's write it themselves?

The way I see it, the `httputil.NewSingleHostReverseProxy(target)` is fairly opaque to the user. If they've **ALREADY** provided a value for `u.Log` (where `u` is an `Upstream`) then that logger should be used for `u.ReverseProxy`.

> Or providing it as an option instead?

Happy to go the route of making it an option, but right now `NewUpstream` doesn't take any options. Members are `Name`, `Log`, `URL`, `ReverseProxy` and `NewUpstream()` takes a `URL` and sets both `URL` and `ReverseProxy`.

reverseproxy/upstream.go
wcharczuk: why do we need the back reference?

dhermes: The caller doesn't have access to `u` when they invoke `NewUpstream()` and the constructor gives no way to pass in `u.Log`. But `rp.ErrorHandler` **should** refer to `u.Log` (which can't be set until `NewUpstream()` exits).

PR #170, opened 2020-01-21
Allowing "extra" key-value pairs in `httptrace` Start().

dhermes:
## PR Summary

 - **Type:** Features


webutil/tracer.go
```diff
@@ -4,14 +4,20 @@ import (
 	"net/http"
 )
 
-// HTTPTracer is a simplified version of `Tracer` intended for a raw
-// `(net/http).Request`. It returns a "new" request the request context may
-// be modified after opening a span.
+// TagKV is a key value pair for a span tag.
+type TagKV struct {
```
dhermes: NOTE: I made this type to avoid taking a dependency on `github.com/opentracing/opentracing-go` here.

stats/tracing/httptrace/tracer.go
```diff
@@ -57,10 +57,15 @@ func StartHTTPSpan(ctx context.Context, tracer opentracing.Tracer, req *http.Req
 
 // Start opens a span and creates a new request with a modified context, based
 // on the span that was opened.
-func (ht httpTracer) Start(req *http.Request) (webutil.HTTPTraceFinisher, *http.Request) {
+func (ht httpTracer) Start(req *http.Request, extraKV ...webutil.TagKV) (webutil.HTTPTraceFinisher, *http.Request) {
```
wcharczuk: can we generalize this to mutate a `StartSpanOptions` type that has a couple other things potentially set on it? i worry about tying us to this level of abstraction

dhermes:
Reviewers: I'm not 100% sold this is "necessary". The original use-case I imagined it would help can be achieved by just setting global tags on the root tracer.

PR #171, opened 2020-01-24
Making ImmediateSchedule.didRun swap atomic.

dhermes:
## PR Summary

 - **Type:** Bugfix


dhermes:
This was made irrelevant by #172.

PR #172, opened 2020-01-24
Fixes `go-sdk/cron` handling of stateful schedules

wcharczuk:
Fixes an issue with a recent refactor of `go-sdk/cron` where we fetched the schedule off the job each execution, meaning that stateful schedules would always return the next run time based on their initial state.


cron/job_scheduler.go
```diff
@@ -20,8 +20,13 @@ func NewJobScheduler(job Job, options ...JobSchedulerOption) *JobScheduler {
 		latch: async.NewLatch(),
 		Job:   job,
 	}
+	// the schedule can be stateful, we must memoize the
```
wcharczuk: this is the thrust of the fix

PR #174, opened 2020-01-30
Expose datadog options in collector

mat285:
## PR Summary

 - **Type:** Features
 - **Issue Link:**

#### Reviewers:

### DIFF
--- a/datadog/collector.go
+++ b/datadog/collector.go
@@ -19,14 +19,13 @@ var (
 )
 
 // New returns a new stats collector from a config.
-func New(cfg Config) (*Collector, error) {
+func New(cfg Config, opts ...statsd.Option) (*Collector, error) {
 	var client *statsd.Client
 	var err error
 	if cfg.BufferedOrDefault() {
-		client, err = statsd.NewBuffered(cfg.GetHost(), cfg.BufferDepthOrDefault())
-	} else {
-		client, err = statsd.New(cfg.GetHost())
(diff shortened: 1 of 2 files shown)


datadog/collector.go
```diff
@@ -19,14 +19,13 @@ var (
 )
 
 // New returns a new stats collector from a config.
-func New(cfg Config) (*Collector, error) {
+func New(cfg Config, opts ...statsd.Option) (*Collector, error) {
```
wcharczuk: tests

datadog/collector_test.go
```diff
@@ -38,17 +40,45 @@ func TestConvertEvent(t *testing.T) {
...
+	assert.Equal("hello", c.client.Namespace)
+
+	cfg.BufferDepth = 10
+	cfg.Buffered = ref.Bool(true)
+
+	c, err = New(cfg, statsd.WithNamespace("hello"))
+	assert.Nil(err)
+	assert.NotNil(c)
+	assert.NotNil(c.client)
+	assert.Equal("hello", c.client.Namespace)
```
mat285: dd doesn't expose the maxmessages option to the client

datadog/collector.go
wcharczuk: one consideration here is leaking that `statsd.` package reference; you will now have to add (2) imports to use this package instead of (1)

mat285: Technically you do not, you only have to import the single package if you aren't using the statsd options. And you already need statsd vendored to vendor this package too

PR #176, opened 2020-02-01
Adds `r2.OptMaxRedirects` which overrides the default of 10

wcharczuk:
Adds an option to set the max redirects, but also adds a utility to check if the error returned on the request was the result of the maximum number of redirects being reached.

### DIFF
--- a/r2/opt_max_redirects.go
+++ b/r2/opt_max_redirects.go
@@ -11,9 +11,8 @@ import (
 
 // OptMaxRedirects tells the http client to only follow a given
 // number of redirects, overriding the standard library default of 10.
-// If a maximum number of redirects is reached, an exception of class `http.ErrUseLastResponse`
-// will be returned, with the redirect history as the exception message.
-// NOTE: that will make the exception message incredibly large.
+// Use the companion helper `ErrIsTooManyRedirects` to test if the returned error
+// from a call indicates the redirect limit was reached.
 func OptMaxRedirects(maxRedirects int) Option {
 	return func(r *Request) error {
 		if r.Client == nil {
--- a/r2/opt_max_redirects.go
+++ b/r2/opt_max_redirects.go
@@ -11,7 +11,7 @@ import (
 )
 
 // OptMaxRedirects tells the http client to only follow a given
-// set of redirects, overriding the standard library default of 10.
+// number of redirects, overriding the standard library default of 10.
 // If a maximum number of redirects is reached, an exception of class `http.ErrUseLastResponse`
 // will be returned, with the redirect history as the exception message.
 // NOTE: that will make the exception message incredibly large.
--- a/r2/opt_max_redirects.go
+++ b/r2/opt_max_redirects.go
@@ -6,9 +6,8 @@ import (
 	"net/url"
 	"strings"
 
-	"github.com/blend/go-sdk/logger"
-
 	"github.com/blend/go-sdk/ex"
+	"github.com/blend/go-sdk/logger"
 )
 
 // OptMaxRedirects tells the http client to only follow a given
--- a/r2/opt_max_redirects.go
+++ b/r2/opt_max_redirects.go
@@ -1,0 +1,51 @@
+package r2
+
+import (
+	"fmt"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/blend/go-sdk/logger"
+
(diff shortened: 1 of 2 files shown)


r2/opt_max_redirects.go
```diff
@@ -0,0 +1,51 @@
+package r2
+
+import (
+	"fmt"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/blend/go-sdk/logger"
+
```
dhermes: Empty line not needed?

r2/opt_max_redirects.go
```diff
@@ -0,0 +1,50 @@
...
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/blend/go-sdk/ex"
+	"github.com/blend/go-sdk/logger"
+)
+
+// OptMaxRedirects tells the http client to only follow a given
+// set of redirects, overriding the standard library default of 10.
```
dhermes: "set of redirects" --> number

r2/opt_max_redirects.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/blend/go-sdk/ex"
+	"github.com/blend/go-sdk/logger"
+)
+
+// OptMaxRedirects tells the http client to only follow a given
+// number of redirects, overriding the standard library default of 10.
+// If a maximum number of redirects is reached, an exception of class `http.ErrUseLastResponse`
+// will be returned, with the redirect history as the exception message.
+// NOTE: that will make the exception message incredibly large.
```
dhermes: "will make" -> "may make"?

r2/opt_max_redirects.go
```diff
@@ -0,0 +1,50 @@
...
+func OptMaxRedirects(maxRedirects int) Option {
+	return func(r *Request) error {
+		if r.Client == nil {
+			r.Client = &http.Client{}
+		}
+		r.Client.CheckRedirect = func(r *http.Request, via []*http.Request) error {
+			if len(via) >= maxRedirects {
+				r = r.WithContext(logger.WithAnnotations(r.Context(), logger.Annotations{
+					"via": urlStrings(via),
+				}))
```
dhermes: How does this help? You'd need to set `r.Context` directly to leave this scope? Or does `WithContext` modify in place? (If yes, no reason to assign to `r`)

r2/opt_max_redirects.go
```diff
@@ -0,0 +1,50 @@
...
+			}
+			return nil
+		}
+		return nil
+	}
+}
+
+// ErrIsTooManyRedirects returns if the error is too many redirects.
+func ErrIsTooManyRedirects(err error) bool {
+	if typed, ok := err.(*url.Error); ok {
```
dhermes: Why the `url.Error` check?

r2/opt_max_redirects.go
wcharczuk: with context returns the modified r, context is set for you by net/http

r2/opt_max_redirects.go
wcharczuk: because the go authors are mean

dhermes: I'm not sure I follow. When would `ErrIsTooManyRedirects()` be used?

r2/opt_max_redirects.go
dhermes: Gotcha: https://github.com/golang/go/blob/go1.13.7/src/net/http/request.go#L360

r2/opt_max_redirects.go
wcharczuk: see the test

r2/opt_max_redirects_test.go
```diff
@@ -0,0 +1,36 @@
...
+		http.Redirect(rw, r, pingURL, http.StatusTemporaryRedirect)
+		return
+	}))
+	defer pong.Close()
+
+	pingURL = ping.URL
+	pongURL = pong.URL
+
+	_, err := New(pingURL, OptMaxRedirects(32)).Discard()
+	assert.True(ErrIsTooManyRedirects(err))
```
wcharczuk: @dhermes usage

dhermes: DERP 🤦‍♂

PR #179, opened 2020-02-01
`go-sdk/cron` simplification & refactor

wcharczuk:
This PR upstreams changes required to make [jobkit](https://github.com/blend/jobkit) take over the bulk of the history tracking features from cron, making the underlying package much simpler.

It also includes changes to other packages like `go-sdk/graceful` to help extend them later.

Important `go-sdk/graceful` usage change note:
- the first user terminate signal `SIGINT` or `SIGTERM` will call `Stop()` on the graceful hosted servers, the second signal will bubble up per normal and exit the process. this prevents zombie graceful processes from hanging on stop.


assert/assert.go
```diff
@@ -62,9 +62,7 @@ func Empty(opts ...Option) *Assertions {
 // New returns a new instance of `Assertions`.
 func New(t *testing.T, opts ...Option) *Assertions {
 	a := Assertions{
-		T:            t,
-		timerAbort:   make(chan bool),
-		timerAborted: make(chan bool),
+		T: t,
```
wcharczuk: NOTE: these timers on assert never really worked because the timer would panic a background thread, letting tests proceed. You *must* use `go test -timeout` to enforce timeouts.

configutil/set.go
```diff
@@ -19,6 +19,23 @@ func SetString(destination *string, sources ...StringSource) error {
 	return nil
 }
 
+// SetStringPtr coalesces a given list of sources into a variable.
+func SetStringPtr(destination **string, sources ...StringSource) error {
```
wcharczuk: GENERICS.

cron/job_config.go
```diff
@@ -4,27 +4,18 @@ import "time"
...
 type JobConfig struct {
-	// Name sets the job name.
-	Name string `json:"name" yaml:"name"`
 	// Disabled determines if the job should be automatically scheduled or not.
 	Disabled *bool `json:"disabled" yaml:"disabled"`
 	// Description is an optional string to describe what the job does.
 	Description string `json:"description" yaml:"description"`
 	// Labels define extra metadata that can be used to filter jobs.
 	Labels map[string]string `json:"labels" yaml:"labels"`
+	// ParameterValues act as default parameters for a given job.
```
wcharczuk: I am not sure `ParameterValues` is the correct name here, everywhere else it's `Parameters` but in jobkit, that means the fully formed input specs (i.e. if the param is a select or a checkbox) vs. the resulting value mapped to the input name.

cron/job_interfaces.go
```diff
@@ -2,32 +2,17 @@ package cron
...
+// Job is an interface types can satisfy to be loaded into the JobManager.
+type Job interface {
+	Name() string
+	Execute(context.Context) error
 }
 
-// LabelsProvider is a type that provides labels.
-type LabelsProvider interface {
-	Labels() map[string]string
+// ConfigProvider is a type that returns a job config.
```
wcharczuk: One aspect of these refactored interfaces that is hard to capture with names is, the schedule provider is called when the job scheduler is created (and cached to prevent stateful schedules getting fouled up). The config provider is called every time the job scheduler needs to check a config option, letting the jobs vary the results over time (such as if they want to be disabled, or change their timeouts etc.)

cron/job_interfaces.go
```diff
@@ -37,105 +22,7 @@ type ScheduleProvider interface {
...
-type HistoryPersistenceEnabledProvider interface {
-	HistoryPersistenceEnabled() bool
-}
-
-// HistoryProvider is a job that can persist and restore its invocation history.
-type HistoryProvider interface {
-	RestoreHistory(context.Context) ([]JobInvocation, error)
-	PersistHistory(context.Context, []JobInvocation) error
+// LifecycleProvider is a job that provides lifecycle hooks.
+type LifecycleProvider interface {
```
wcharczuk: This is a big change.

We instead of sniffing 8 billion interfaces just have the job return a struct with delegate fields, that if not nil are assumed callable. This simplifies code that extends jobs with extra stuff on the lifecycle steps (i.e. jobkit) but also fully enumerates all the lifecycle hooks so you know what ones are available as a user.

cron/job_invocation.go
```diff
@@ -34,85 +24,27 @@ func NewJobInvocationID() string {
...
-	Context context.Context
-	Cancel  context.CancelFunc
-	Done    chan struct{}
-}
+	Started  time.Time `json:"started"`
+	Complete time.Time `json:"complete"`
+	Err      error     `json:"err"`
 
-// MarshalJSON marshals the invocation as json.
-func (ji JobInvocation) MarshalJSON() ([]byte, error) {
```
wcharczuk: JSON serialization is not something everyone needs for invocation history, i've moved these methods to the jobkit.JobInvocation type which also includes captured job output etc.

cron/job_invocation.go
```diff
@@ -34,85 +24,27 @@ func NewJobInvocationID() string {
...
-	ji.Elapsed = values.Elapsed
-	ji.State = values.State
-	if values.Error != "" {
-		ji.Err = errors.New(values.Error)
-	}
-	ji.Output = new(bufferutil.Buffer)
-	if err := json.Unmarshal([]byte(values.Output), ji.Output); err != nil {
-		return ex.New(err)
+// Elapsed returns the elapsed time for the invocation.
+func (ji JobInvocation) Elapsed() time.Duration {
```
wcharczuk: TBD if it's more ergonomic to just recalculate elapsed time each access, but it seemed duplicative to have elapsed time AND started + completed timestamps.

cron/job_manager.go
```diff
@@ -198,96 +259,16 @@ func (jm *JobManager) CancelJob(jobName string) (err error) {
...
-		}
 		return JobManagerStateRunning
 	} else if jm.Latch.IsStopped() {
 		return JobManagerStateStopped
 	}
 	return JobManagerStateUnknown
 }
-
-// Status returns a status object.
-func (jm *JobManager) Status() JobManagerStatus {
```
wcharczuk: This moved to jobkit

cron/job_scheduler_test.go
```diff
@@ -2,100 +2,22 @@ package cron
...
 	"github.com/blend/go-sdk/graceful"
-	"github.com/blend/go-sdk/ref"
-	"github.com/blend/go-sdk/uuid"
 )
 
 var (
 	_ graceful.Graceful = (*JobScheduler)(nil)
 )
 
-func TestJobSchedulerCullHistoryMaxAge(t *testing.T) {
```
wcharczuk: history stuff moved to jobkit

cron/job_scheduler_test.go
```diff
@@ -196,43 +59,52 @@ func TestJobSchedulerLabels(t *testing.T) {
 	assert.Equal("not-test", labels["name"])
 	assert.Equal("bar", labels["foo"])
 	assert.Equal("wuzz", labels["fuzz"])
-	assert.Equal(JobInvocationStateComplete, labels["last"])
+	assert.Equal(JobInvocationStatusSuccess, labels["last"])
 }
 
-func TestJobSchedulerStats(t *testing.T) {
+func TestJobSchedulerJobParameterValues(t *testing.T) {
```
wcharczuk: We now coalesce parameter values from the job config, and the invocation context, this lets you have parameterful jobs that don't have to do their own coalesce every invocation

graceful/shutdown_by_signal.go
```diff
@@ -0,0 +1,80 @@
...
+			<-shouldShutdown // tell the hosted process to stop "gracefully"
+			if err := instance.Stop(); err != nil {
+				errors <- err
+			}
+			return
+		}(hostedInstance)
+	}
+
+	select {
+	case <-options.ShutdownSignal: // if we've issued a shutdown, wait for the server to exit
```
wcharczuk: BIG CHANGE.

The first c-c will tell the graceful servers to stop. A second user c-c will kill the process.

PR #184, opened 2020-02-04
Adding `jsonutil` package.

dhermes:
## PR Summary

 - **Type:** Features


jsonutil/flatten.go
```diff
@@ -0,0 +1,12 @@
...
+
+// Flatten takes a map and returns an array of the values,
+// discarding the map keys.
+func Flatten(obj interface{}) (values []interface{}) {
+	if typed, ok := obj.(map[string]interface{}); ok {
+		for _, value := range typed {
+			values = append(values, value)
+		}
+	}
+	return
```
dhermes: I don't know if this is intended @wcharczuk but this will be non-deterministic for a given `map` literal. `map` ordering is intentionally random.

dhermes:
Good to know! Thanks.

PR #186, opened 2020-02-05
Making `go-sdk/db.connection.begin` less awkward to use

wcharczuk:
Also fixes an issue in autoflush buffer with adding items before it's started.


db/connection.go
```diff
@@ -108,20 +108,23 @@ func (dbc *Connection) Open() error {
...
+func (dbc *Connection) BeginContext(context context.Context, opts ...func(*sql.TxOptions)) (*sql.Tx, error) {
 	if dbc.Connection == nil {
 		return nil, ex.New(ErrConnectionClosed)
 	}
-	if len(opts) > 0 {
-		tx, err := dbc.Connection.BeginTx(context, opts[0])
-		return tx, Error(err)
+	var txOptions sql.TxOptions
+	for _, opt := range opts {
+		opt(&txOptions)
```
dhermes: Big 👍 here. (I am not a big fan of the "0 or 1" use of variadic inputs.)

PR #188, opened 2020-03-18
Add datadog address field to support unix domain socket

chotiwat:
## PR Summary

This PR adds an `address` field (env: `DATADOG_ADDRESS`) that allows the user to send metrics to a datadog agent either via UDP at `host:port` or unix domain socket at `unix:///path/to/socket`.

 - **Type:** Features
 - **Issue Link:** N/A

#### Reviewers:


dvdliao:
strange how sentry pkg test failing

wcharczuk:
i have to merge my GRPC tracing stuff, let me do that first.

datadog/config.go
```diff
@@ -67,6 +69,14 @@ func (c Config) TracingEnabledOrDefault() bool {
 	return DefaultTracingEnabled
 }
 
+// GetAddress returns the datadog collector address string.
+func (c Config) GetAddress() string {
+	if c.Address != "" {
+		return c.Address
+	}
+	return c.GetHost()
```
wcharczuk: this is conflating the two fields, we should join the host and the port to create a fully formed URI, you will also probably need some helpers for sniffing for `unix://` on the prefix of the address but will depend on how the dd stuff works

datadog/config.go
```diff
@@ -14,6 +14,8 @@ const (
 
 // Config is the datadog config.
 type Config struct {
+	// Address is the address of the datadog collector in the form of "hostname:port" or "unix:///path/to/socket"
+	Address string `json:"address,omitempty" yaml:"address,omitempty" env:"DATADOG_ADDRESS"`
```
dhermes: Why wouldn't it be in the form of `{protocol}://{hostname}:{port}`?

datadog/config.go
dhermes: I'd imagine `statsd.New()` is following Postel's law and being lenient if the caller passes host+port without a protocol.

datadog/config.go
wcharczuk: for networked collectors the protocol is UDP so it's inferred and not part of a proto in a uri.

dhermes: Ah duh, me missing the whole point of this PR 😁

PR #189, opened 2020-03-24
Add Flush to stats.Collector interface

chotiwat:
## PR Summary

This PR adds `Flush() error` to `stats.Collector` interface.

 - **Type:** Internal
 - **Issue Link:** N/A

#### Reviewers:

### DIFF
--- a/stats/mock_collector.go
+++ b/stats/mock_collector.go
@@ -15,8 +15,9 @@ var (
 // NewMockCollector returns a new mock collector.
 func NewMockCollector() *MockCollector {
 	return &MockCollector{
-		Events: make(chan MockMetric, 32),
-		Errors: make(chan error, 32),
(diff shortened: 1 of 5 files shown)


stats/mock_collector.go
```diff
@@ -84,6 +84,14 @@ func (mc MockCollector) TimeInMilliseconds(name string, value time.Duration, tag
 	return nil
 }
 
+// Flush does nothing on a MockCollector.
+func (mc MockCollector) Flush() error {
+	if len(mc.Errors) > 0 {
+		return <-mc.Errors
```
wcharczuk: it doesn't follow that the `flush` will return errors that were "collected" when doing metrics submissions, so we may not want to add this here.

i would add a new channel, `mc.FlushErrors` that would handle those use cases.

PR #197, opened 2020-04-24
Adds additional types for configutil resolvers.

wcharczuk:


configutil/duration_ptr.go
```diff
@@ -0,0 +1,22 @@
+package configutil
+
+import "time"
+
+// DurationPtr returns a DurationSource for a given duration pointer.
+func DurationPtr(value *time.Duration) DurationSource {
```
dhermes, line 6: Preference on `Ptr` vs. `Pointer`? IMO abbreviation shouldn't be used here

wcharczuk: Ptr is used in the stdlib

PR #200, opened 2020-06-16
Adding `retry` and `envoy` packages.

dhermes:
## PR Summary

 - **Type:** Features

#### Reviewers:


envoy/wait_test.go
```diff
@@ -0,0 +1,241 @@
...
+		"[debug] Envoy is ready",
+		"",
+	}, "\n")
+	it.Equal(expected, string(logBuffer.Bytes()))
+}
+
+// getMinorVersion attempts to extract the minor (`X`) version in a version
+// string `go1.X` or `go1.X.Y`, e.g. `go1.14.4`. If the version is not of
+// this form, a negative number will be returned.
+func getMinorVersion() int {
```
wcharczuk, line 122: i would probably not do this, is there a way to make the test more permissive of quotes?

dhermes: Swapped out for a regex.

envoy/COVERAGE
```diff
@@ -0,0 +1 @@
+100.00
```
wcharczuk, line 1: SUSPECT.

PR #202, opened 2020-06-17
Using `uint` for `attempts` in `retry` package.

dhermes:
## PR Summary

This prevents a runtime panic with a negative bitshift in `1 << attempts`.

 - **Type:** Bugfix


dhermes:
With

```go
main.go
package main
import (
	"fmt"
)
func Shift(attempt int) {
	fmt.Println(1 << attempt)
}
func main() {
	Shift(10)
	Shift(-3)
}
```

there is a difference between 1.12 and 1.13

```
$ docker run --rm --interactive --tty --volume $(pwd):/var/dhermes golang:1.12.17-alpine go run /var/dhermes/main.go
# command-line-arguments
/var/dhermes/main.go:8:16: invalid operation: 1 << attempt (shift count type int, must be unsigned integer)
$
$ docker run --rm --interactive --tty --volume $(pwd):/var/dhermes golang:1.13.12-alpine go run /var/dhermes/main.go
1024
panic: runtime error: negative shift amount
goroutine 1 [running]:
main.Shift(...)
        /var/dhermes/main.go:8
main.main()
        /var/dhermes/main.go:13 +0x86
exit status 2
```

PR #204, opened 2020-07-16
Adding `spiffeutil` package for parsing.

dhermes:
## PR Summary

 - **Type:** Features

#### Reviewers:


spiffeutil/parse.go
```diff
@@ -0,0 +1,32 @@
...
+
+	// ErrInvalidURI is the class of error returned when parsing SPIFFE URI fails
+	ErrInvalidURI = ex.Class("Invalid SPIFFE URI")
+)
+
+// Parse consumes a SPIFFE URI and splits out the trust domain and workload
+// identifier. For example in `spiffe://cluster.local/ns/blend/sa/quasar`
+// the trust domain is `cluster.local` and the workload identifier is
+// `ns/blend/sa/quasar`.
+func Parse(uri string) (string, string, error) {
```
wcharczuk, line 20: can we use named outputs here or even better use a struct output

dhermes: Sure. Will go with the struct.

PR #206, opened 2020-07-17
Adding `envoyutil.ExtractClientIdentity()`.

dhermes:
## PR Summary

 - **Type:** Features

#### Reviewers:

### DIFF
--- a/envoyutil/identity.go
+++ b/envoyutil/identity.go
@@ -57,37 +57,33 @@ type VerifyXFCC func(xfcc XFCCElement, xfccValue string) (err *XFCCError)
 // does so by requiring the XFCC header to present and valid and then passing
 // the parsed XFCC header along to some verifiers (e.g. to verify the server
 // identity) as well as to an extractor (for the client identity).
-func ExtractClientIdentity(req *http.Request, efx ExtractFromXFCC, verifiers ...VerifyXFCC) (clientIdentity string, err *XFCCError) {
+func ExtractClientIdentity(req *http.Request, efx ExtractFromXFCC, verifiers ...VerifyXFCC) (string, *XFCCError) {
 	if efx == nil {
-		err = &XFCCError{Class: ErrMissingExtractFunction}
-		return
+		return "", &XFCCError{Class: ErrMissingExtractFunction}
 	}
 
 	// Early exit if XFCC header is not present.
 	xfccValue := req.Header.Get(HeaderXFCC)
 	if xfccValue == "" {
-		err = &XFCCError{Class: ErrMissingXFCC}
-		return
+		return "", &XFCCError{Class: ErrMissingXFCC}
 	}
 
 	// Early exit if XFCC header is invalid, or has zero or multiple elements.
 	xfccElements, parseErr := ParseXFCC(xfccValue)
 	if parseErr != nil || len(xfccElements) != 1 {
-		err = &XFCCError{Class: ErrInvalidXFCC, XFCC: xfccValue}
-		return
+		return "", &XFCCError{Class: ErrInvalidXFCC, XFCC: xfccValue}
 	}
 	xfcc := xfccElements[0]
 
 	// Run all verifiers on the parsed `xfcc`.
 	for _, verifier := range verifiers {
 		if verifier == nil {
-			err = &XFCCError{Class: ErrVerifierNil, XFCC: xfccValue}
-			return
+			return "", &XFCCError{Class: ErrVerifierNil, XFCC: xfccValue}
 		}
 
-		err = verifier(xfcc, xfccValue)
+		err := verifier(xfcc, xfccValue)
 		if err != nil {
-			return
+			return "", err
 		}
 	}
 
--- a/envoyutil/identity.go
+++ b/envoyutil/identity.go
@@ -10,15 +10,16 @@ const (
 	// ErrNilInterface is the error returned when a user error results in `nil`
 	// being passed as a function interface type.
 	ErrNilInterface = ex.Class("User error: nil interface provided")
-	// ErrMissingExtractFunction is the error returned when the "extract client
-	// identity" function is `nil` or not provided.
-	ErrMissingExtractFunction = ex.Class("Missing client identity extraction function")
-	// ErrVerifierNil is the error returned when a provided verifier is `nil`.
-	ErrVerifierNil = ex.Class("XFCC verifier must not be `nil`")
 	// ErrMissingXFCC is the error returned when XFCC is missing
 	ErrMissingXFCC = ex.Class("Missing X-Forwarded-Client-Cert header")
 	// ErrInvalidXFCC is the error returned when XFCC is invalid
 	ErrInvalidXFCC = ex.Class("Invalid X-Forwarded-Client-Cert header")
+
+	// MissingExtractFunction is the message used when the "extract client
+	// identity" function is `nil` or not provided.
+	MissingExtractFunction = "Missing client identity extraction function"
+	// VerifierNil is the message prefix used when a provided verifier is `nil`.
+	VerifierNil = "XFCC verifier must not be `nil`"
 )
 
 // NOTE: Ensure
@@ -62,7 +63,7 @@ type VerifyXFCC func(xfcc XFCCElement, xfccValue string) (err *XFCCError)
 // identity) as well as to an extractor (for the client identity).
 func ExtractClientIdentity(req *http.Request, efx ExtractFromXFCC, verifiers ...VerifyXFCC) (string, error) {
 	if efx == nil {
-		return "", ex.New(ErrNilInterface, ex.OptMessage(ErrMissingExtractFunction))
+		return "", ex.New(ErrNilInterface, ex.OptMessage(MissingExtractFunction))
 	}
 
 	// Early exit if XFCC header is not present.
@@ -83,7 +84,7 @@ func ExtractClientIdentity(req *http.Request, efx ExtractFromXFCC, verifiers ...
 		if verifier == nil {
 			return "", ex.New(
 				ErrNilInterface,
-				ex.OptMessagef("%s; XFCC: %q", ErrVerifierNil, xfccValue),
+				ex.OptMessagef("%s; XFCC: %q", VerifierNil, xfccValue),
 			)
 		}
 
--- a/envoyutil/identity.go
+++ b/envoyutil/identity.go
@@ -22,31 +22,6 @@ const (
 	VerifierNil = "XFCC verifier must not be `nil`"
 )
 
-// NOTE: Ensure
-//       - `XFCCError` satisfies `error`.
-var (
-	_ error = (*XFCCError)(nil)
-)
-
-// XFCCError contains metadata about an XFCC header that could not
-// be parsed. This is intended to be used as the body of a 401 Unauthorized
-// response.
-type XFCCError struct {
-	// Class can be used to uniquely identify the type of the error.
-	Class ex.Class `json:"class" xml:"class"`
-	// XFCC contains the XFCC header value that could not be parsed or was
-	// invalid in some way.
-	XFCC string `json:"xfcc,omitempty" xml:"xfcc,omitempty"`
-	// Metadata contains extra information relevant to a specific failure.
-	Metadata interface{} `json:"metadata,omitempty" xml:"metadata,omitempty"`
-}
-
-// Error satisfies the `error` interface. It is intended to be a unique
-// identifier for the error.
-func (xe *XFCCError) Error() string {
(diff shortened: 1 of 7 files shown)


envoyutil/identity.go
```diff
@@ -0,0 +1,78 @@
...
+// addition to the client identity.
+type VerifyXFCC func(xfcc XFCCElement, xfccValue string) (errResponse *InvalidXFCCResponse)
+
+// ExtractClientIdentity enables extracting client identity from a request. It
+// does so by requiring the XFCC header to present and valid and then passing
+// the parsed XFCC header along to some verifiers (e.g. to verify the server
+// identity) as well as to an extractor (for the client identity).
+func ExtractClientIdentity(ctx *web.Ctx, efx ExtractFromXFCC, verifiers ...VerifyXFCC) (clientIdentity string, errResponse *InvalidXFCCResponse) {
+	if efx == nil {
+		errResponse = &InvalidXFCCResponse{Message: ErrMissingExtractFunction}
```
dhermes, line 44: I considered just allowing this to `panic()`. One thing I don't like is that `InvalidXFCCResponse{}` is in some sense intended to be returned as an HTTP 401 response body, but this is like "hey user error".

envoyutil/identity.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil || len(xfccElements) != 1 {
+		errResponse = &InvalidXFCCResponse{Message: ErrInvalidXFCC, XFCC: xfccValue}
+		return
+	}
+	xfcc := xfccElements[0]
+
+	// Run all verifiers on the parsed `xfcc`.
+	for _, verifier := range verifiers {
+		if verifier == nil {
+			errResponse = &InvalidXFCCResponse{Message: ErrVerifierNil, XFCC: xfccValue}
```
dhermes, line 66: Ditto here on user error vs. a real parsing failure.

envoyutil/identity.go
```diff
@@ -0,0 +1,78 @@
...
+
+		errResponse = verifier(xfcc, xfccValue)
+		if errResponse != nil {
+			return
+		}
+	}
+
+	// Do final extraction.
+	return efx(xfcc, xfccValue)
+}
```
dhermes, line 78: I am in the process of actually implementing the functions that satisfy `ExtractFromXFCC` and `VerifyXFCC` (they will be SPIFFE aware, among other things).

I can add them in this PR if deemed necessary but I wanted to get this out there because it's big enough to review and is self-contained.

dhermes: https://github.com/dhermes/go-sdk/compare/client-identity...client-identity-add-processors <-- this change is in the wings as well

envoyutil/identity.go
```diff
@@ -0,0 +1,78 @@
...
+// VerifyXFCC is an "extra" verifier for an XFCC, for example if the server
+// identity (from the `By` field in an XFCC element) should be verified in
+// addition to the client identity.
+type VerifyXFCC func(xfcc XFCCElement, xfccValue string) (errResponse *InvalidXFCCResponse)
+
+// ExtractClientIdentity enables extracting client identity from a request. It
+// does so by requiring the XFCC header to present and valid and then passing
+// the parsed XFCC header along to some verifiers (e.g. to verify the server
+// identity) as well as to an extractor (for the client identity).
+func ExtractClientIdentity(ctx *web.Ctx, efx ExtractFromXFCC, verifiers ...VerifyXFCC) (clientIdentity string, errResponse *InvalidXFCCResponse) {
```
wcharczuk, line 42: can we have this operate on `http.Request` like we discussed

wcharczuk: for the response, there are a couple ways to do this, the typical one is with typed errors or error classes

dhermes: > can we have this operate on `http.Request` like we discussed

Absolutely. It slipped my mind, thanks for the reminder.

> the typical one is with typed errors or error classes

I considered making `errResponse` satisfy `error`, but ended up punting on it. What would you want to return in the string?

envoyutil/identity.go
```diff
@@ -0,0 +1,94 @@
...
+// VerifyXFCC is an "extra" verifier for an XFCC, for example if the server
+// identity (from the `By` field in an XFCC element) should be verified in
+// addition to the client identity.
+type VerifyXFCC func(xfcc XFCCElement, xfccValue string) (err *XFCCError)
+
+// ExtractClientIdentity enables extracting client identity from a request. It
+// does so by requiring the XFCC header to present and valid and then passing
+// the parsed XFCC header along to some verifiers (e.g. to verify the server
+// identity) as well as to an extractor (for the client identity).
+func ExtractClientIdentity(req *http.Request, efx ExtractFromXFCC, verifiers ...VerifyXFCC) (clientIdentity string, err *XFCCError) {
```
dhermes, line 58: I'm still on the fence about `err *XFCCError` vs. `err error` here.

envoyutil/identity.go
```diff
@@ -0,0 +1,94 @@
...
+	if efx == nil {
+		err = &XFCCError{Class: ErrMissingExtractFunction}
```
dhermes, line 60: I considered just allowing this to `panic`. One thing I don't like is that `XFCCError{}` is in some sense intended to be returned as an HTTP 401 response body, whereas this error is like "hey, user error".

(I suppose that could be an argument for using `ex.New()` here instead.)

envoyutil/identity.go
```diff
@@ -0,0 +1,94 @@
...
+	if parseErr != nil || len(xfccElements) != 1 {
+		err = &XFCCError{Class: ErrInvalidXFCC, XFCC: xfccValue}
+		return
+	}
+	xfcc := xfccElements[0]
+
+	// Run all verifiers on the parsed `xfcc`.
+	for _, verifier := range verifiers {
+		if verifier == nil {
+			err = &XFCCError{Class: ErrVerifierNil, XFCC: xfccValue}
```
dhermes, line 82: Ditto here on user error vs. a real parsing failure.

envoyutil/identity.go
```diff
@@ -0,0 +1,94 @@
+package envoyutil
+
+import (
+	"net/http"
+)
+
+const (
+	// ErrMissingExtractFunction is the error returned when the "extract client
+	// identity" function is `nil` or not provided.
+	ErrMissingExtractFunction = "Missing client identity extraction function"
```
wcharczuk, line 10: this is named `Err` but is not typed `error`, can you either use `ex.Class` or another string wrapper that implements error, or rename the constants

envoyutil/identity_test.go
```diff
@@ -0,0 +1,104 @@
...
+	assert := sdkAssert.New(t)
+
+	type testCase struct {
+		XFCC           string
+		ClientIdentity string
+		Class          string
+		Extract        envoyutil.ExtractFromXFCC
+		Verifiers      []envoyutil.VerifyXFCC
+	}
+	testCases := []testCase{
```
wcharczuk, line 39: (NOT REQUIRED)

you can use anonymous struct definitions and preallocated arrays for this:

```
[...]struct{
    XFCC string
    ClientIdentity string
    Class string
    ...
} {
    { XFCC: "" ... },
}
```

etc.

dhermes: Yup, am aware. I prefer having a wee bit of separation between the struct definition and the test cases, but it's 100% taste / not based on any objective truth.

envoyutil/identity.go
dhermes: I went with making them `ex.Class` values.

envoyutil/identity.go
wcharczuk: for the return type? i would do `error` if you intend it to be used in `if err != nil {` like blocks

envoyutil/identity.go
wcharczuk: yeah i would have a way to clearly identify if a returned error is a user error. maybe a helper?

envoyutil/identity.go
dhermes: Done

envoyutil/identity.go
dhermes: Added a class `ErrNilInterface` and a helper `IsUserError()`.

envoyutil/identity.go
```diff
@@ -0,0 +1,106 @@
...
+	// invalid in some way.
+	XFCC string `json:"xfcc,omitempty" xml:"xfcc,omitempty"`
+	// Metadata contains extra information relevant to a specific failure.
+	Metadata interface{} `json:"metadata,omitempty" xml:"metadata,omitempty"`
+}
+
+// Error satisfies the `error` interface. It is intended to be a unique
+// identifier for the error.
+func (xe *XFCCError) Error() string {
+	return string(xe.Class)
```
wcharczuk, line 47: also `xe.Class.String()`

envoyutil/identity.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// IsUserError determines if an error is a user error (vs. a legitimate input
+// error) in this package. A user error happens when a `nil` interface is
+// provided in a place where a non-`nil` value is required.
+func IsUserError(err error) bool {
+	return ex.Is(err, ErrNilInterface)
```
wcharczuk, line 54: hrm. if this works, it works, but afaik it's kind of convoluted, basically `ex.Is(...)` does a bunch of stuff to check the contents of the error: 
- is it an `ex.Ex`? use ex.Class == class
- is it an `*ex.Ex`? use ex.Class == class
- is it an `error`? use .Error() == class

this will fail for both the first two, but (should?) work for the 3rd. it relies on the `.Error` function returning a string for the class you're comparing to

wcharczuk: might want to just do the type coercion yourself and check the xe.Class yourself

envoyutil/identity.go
dhermes: Done

envoyutil/identity.go
dhermes: Done

envoyutil/xfcc_test.go
```diff
@@ -60,6 +60,33 @@ JmUY+1YGMn7qbeHTma33N28Ec7hK+WByul746Nro
...
+	for _, tc := range testCases {
+		xe := envoyutil.XFCCElement{By: tc.By, URI: tc.URI}
+		assert.Equal(tc.Expected, xe.String())
+		parsed, err := envoyutil.ParseXFCCElement(xe.String())
+		assert.Nil(err)
+		assert.Equal(xe, parsed)
+	}
+
+	// NOTE: For quoting and unquoting, `ParseXFCCElement()` is not fully an
+	//       inverse to `XFCCElement.String()`.
```
dhermes, line 85: This is essentially a bug in `ParseXFCCElement()`; I'll attempt to address it in a different PR.

envoyutil/xfcc_errors.go
```diff
@@ -0,0 +1,88 @@
...
+	_ error = (*XFCCValidationError)(nil)
+	_ error = (*XFCCFatalError)(nil)
+)
+
+// XFCCExtractionError contains metadata about an XFCC header that could not
+// be parsed or extracted. This is intended to be used as the body of a 401
+// Unauthorized response.
+type XFCCExtractionError struct {
+	// Class can be used to uniquely identify the type of the error.
+	Class ex.Class `json:"class" xml:"class"`
```
dhermes, line 22: @wcharczuk Do you still think an `ex.Class` is necessary here vs. just a `string` called `message` (in the JSON)? Now that the `XFCCExtractionError`, `XFCCValidationError` and `XFCCFatalError` types can be type-switched on, the `ex.Class` seems not needed.

envoyutil/identity.go
```diff
@@ -0,0 +1,67 @@
...
+	// identity" function is `nil` or not provided.
+	MissingExtractFunction = ex.Class("Missing client identity extraction function")
+	// VerifierNil is the message prefix used when a provided verifier is `nil`.
+	VerifierNil = ex.Class("XFCC verifier must not be `nil`")
+)
+
+// ClientIdentityProvider is a function to extra the client identity from a
+// parsed XFCC header. For example, client identity could be determined from the
+// SPIFFE URI in the `URI` field in an XFCC element.
+type ClientIdentityProvider func(xfcc XFCCElement) (clientIdentity string, err error)
```
dhermes, line 24: Note this still makes the assumption that this is a single element vs. a slice `XFCC`.

I don’t feel clear on whether i want these `func` interfaces to take an `XFCC` or `XFCCElement` but am leaning towards `XFCCElement`. It just makes it painful if it’s an `XFCC` because then they can’t coordinate on "choosing" which one, or each extractor / verifier would need to fail if `len != 1`.

PR #207, opened 2020-07-23
Adding `spiffeutil.ParseKubernetesWorkloadID()`.

dhermes:
## PR Summary

 - **Type:** Features

#### Reviewers:

### DIFF
--- a/spiffeutil/parse_test.go
+++ b/spiffeutil/parse_test.go
@@ -3,13 +3,14 @@ package spiffeutil_test
 import (
 	"testing"
 
-	"github.com/blend/go-sdk/assert"
+	sdkAssert "github.com/blend/go-sdk/assert"
 	"github.com/blend/go-sdk/ex"
+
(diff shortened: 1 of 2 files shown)


spiffeutil/parse_test.go
```diff
@@ -3,13 +3,14 @@ package spiffeutil_test
 import (
 	"testing"
 
-	"github.com/blend/go-sdk/assert"
+	sdkAssert "github.com/blend/go-sdk/assert"
```
wcharczuk, line 6: (NON ACTION)

the way that graham does this is he does `it := assert.New(t)` which then lets things read `it.NotNil(foo)`. this is fine though

dhermes: Yeah I have also used that pattern. Our team decided that they liked `assert.Equal(a, b)` but we added some linting to ban shadowing of imports, so that was the "style" we agreed on. But I'm very open to other approaches in other codebases.

PR #208, opened 2020-07-23
Adding a `ClientIdentityFromSPIFFE()` helper.

dhermes:
## PR Summary

 - **Type:** Features

#### Reviewers:

### DIFF
--- a/envoyutil/client_identity_processor.go
+++ b/envoyutil/client_identity_processor.go
@@ -31,19 +31,19 @@ func OptDeniedTrustDomains(trustDomains ...string) ClientIdentityProcessorOption
 }
 
 // OptAllowedClientIdentities adds allowed client identities to the processor.
-func OptAllowedClientIdentities(clientIDs ...string) ClientIdentityProcessorOption {
+func OptAllowedClientIdentities(clientIdentities ...string) ClientIdentityProcessorOption {
 	return func(cip *ClientIdentityProcessor) {
 		cip.AllowedClientIdentities = cip.AllowedClientIdentities.Union(
-			collections.NewSetOfString(clientIDs...),
+			collections.NewSetOfString(clientIdentities...),
 		)
 	}
 }
 
 // OptDeniedClientIdentities adds denied client identities to the processor.
-func OptDeniedClientIdentities(clientIDs ...string) ClientIdentityProcessorOption {
+func OptDeniedClientIdentities(clientIdentities ...string) ClientIdentityProcessorOption {
 	return func(cip *ClientIdentityProcessor) {
 		cip.DeniedClientIdentities = cip.DeniedClientIdentities.Union(
-			collections.NewSetOfString(clientIDs...),
+			collections.NewSetOfString(clientIdentities...),
 		)
 	}
 }
@@ -112,18 +112,18 @@ func (cip ClientIdentityProcessor) ClientIdentityProvider(xfcc XFCCElement) (str
 		return "", err
 	}
 
-	clientID, err := cip.formatClientID(xfcc, pu)
+	clientIdentity, err := cip.formatClientIdentity(xfcc, pu)
 	if err != nil {
 		return "", err
 	}
 
-	if err := cip.ProcessAllowedClientIdentities(xfcc, clientID); err != nil {
+	if err := cip.ProcessAllowedClientIdentities(xfcc, clientIdentity); err != nil {
 		return "", err
 	}
-	if err := cip.ProcessDeniedClientIdentities(xfcc, clientID); err != nil {
+	if err := cip.ProcessDeniedClientIdentities(xfcc, clientIdentity); err != nil {
 		return "", err
 	}
-	return clientID, nil
+	return clientIdentity, nil
 }
 
 // ProcessAllowedTrustDomains returns an error if an allow list is configured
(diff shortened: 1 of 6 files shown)


envoyutil/identity.go
```diff
@@ -68,3 +75,64 @@ func ExtractAndVerifyClientIdentity(req *http.Request, cip ClientIdentityProvide
...
+// ClientIdentityFromSPIFFE produces a function satisfying `ClientIdentityProvider`.
+//
+// This function assumes the client identity is in the `URI` field and that field
+// is a SPIFFE URI. Further, it assumes the workload identifier identifies a
+// Kubernetes service account, of the form  `ns/{namespace}/sa/{serviceAccount}`.
+// After parsing, the returned value is of the form `{namespace}.{serviceAccount}`.
+//
+// Additionally, it takes a variadic input of `denied` client identities that
+// should not pass validation.
+func ClientIdentityFromSPIFFE(trustDomain string, denied ...string) ClientIdentityProvider {
```
dhermes, line 112: @wcharczuk I am a bit torn on whether `denied` should be included in `ClientIdentityFromSPIFFE()` or if there should be two functions exposed. Open to suggestions.

envoyutil/identity.go
```diff
@@ -68,3 +75,64 @@ func ExtractAndVerifyClientIdentity(req *http.Request, cip ClientIdentityProvide
...
+	// NOTE: The `pu == nil` check is redundant, we expect `spiffeutil.Parse()`
+	//       not to violate the invariant that `pu != nil` when `err == nil`.
+	if err != nil || pu == nil {
+		return "", &XFCCExtractionError{
+			Class: ErrInvalidClientIdentity,
+			XFCC:  xfcc.String(),
+		}
+	}
+
+	if pu.TrustDomain != trustDomain {
```
wcharczuk, line 93: are there any casing issues here? do we want to use `strings.EqualFold`?

envoyutil/identity.go
```diff
@@ -68,3 +75,64 @@ func ExtractAndVerifyClientIdentity(req *http.Request, cip ClientIdentityProvide
 	// Do final extraction.
 	return cip(xfcc)
 }
+
+// ClientWorkloadID determines the SPIFFE workload identifier for the client.
+// This function assumes the client identity is in the `URI` field and that field
+// is a SPIFFE URI.
+func ClientWorkloadID(trustDomain string, xfcc XFCCElement) (string, error) {
```
wcharczuk, line 82: I would add something in the comments to the effect of that the trust domain will be validated.

wcharczuk, line 82: another comment here is this function is a bit more involved than the noun `ClientWorkloadID` would indicate.

I would have this function take an options list that mutates a `WorkloadProcessor` or similar, where you can add `AllowList` domains, `DenyList` domains, and potentially other custom validators, a method on this type would then be returned as the delegate return.

envoyutil/client_identity_processor.go
```diff
@@ -0,0 +1,208 @@
...
+// ClientIdentityFormatter describes functions that will produce a client
+// identity string from a parsed SPIFFE URI.
+type ClientIdentityFormatter = func(XFCCElement, *spiffeutil.ParsedURI) (string, error)
+
+// ClientIdentityProcessor provides configurable fields that can be used to
+// help validate a parsed SPIFFE URI and produce and validate a client identity
+// from a parsed SPIFFE URI.
+type ClientIdentityProcessor struct {
+	AllowedTrustDomains     []string
+	DeniedTrustDomains      []string
```
dhermes, line 83: @wcharczuk When you introduced `AllowedTrustDomains` and `DeniedTrustDomains` I wonder if you were mixing up my usage of

```go
if pu.TrustDomain != trustDomain {
	// ...
}
```

and

```go
if deniedSet.Contains(clientIdentity) {
	// ...
}
```

(See original commit: https://github.com/blend/go-sdk/pull/208/commits/e75963212222cf7844b03b12c207e9086b42fabc)

Based on the usage of

```go
		cip := envoyutil.ClientIdentityFromSPIFFE(
			envoyutil.OptAllowedTrustDomains(tc.TrustDomain),
			envoyutil.OptDeniedTrustDomains(tc.Denied...),
		)
```

in the unit tests in https://github.com/blend/go-sdk/pull/209, it seems you had conflated the two concepts by understandable review mistake.

----

IMO having `AllowedTrustDomains` and `DeniedTrustDomains` is likely not necessary. Pinning to a single trust domain for a k8s cluster (i.e. `cluster.local`) and then doing case-insensitive matching on it should be sufficient.

Having 2 allow lists and 2 deny lists is a bit of overhead for this type, but if you think it's worth it, we can leave them as-is.

envoyutil/client_identity_processor.go
```diff
@@ -0,0 +1,208 @@
...
+		Metadata: map[string]string{
+			"clientID": clientID,
+		},
+	}
+}
+
+// ProcessDeniedClientIdentities returns an error if a denied list is configured
+// and a client ID matches any elements in the list.
+func (cip ClientIdentityProcessor) ProcessDeniedClientIdentities(xfcc XFCCElement, clientID string) error {
+	if cip.DeniedClientIdentities.Len() == 0 {
```
dhermes, line 186: Not sure if it's worth it to call `.Len()` vs. just using `len(cip.DeniedClientIdentities)`. Ditto for `len(cip.AllowedClientIdentities)`.

envoyutil/xfcc.go
```diff
@@ -236,7 +236,7 @@ func ParseXFCC(header string) (XFCC, error) {
 		}
 
 		// Increment the index for the next iteration. (Note that branches of the
-		// `switch` statement may have already incremented `i` as well.)
+		// `switch` statement may have already incremented `xp.Index` as well.)
```
dhermes, line 239: I missed this in #212.

dhermes:
Closed in favor of #213.

PR #211, opened 2020-07-28
Adding back support for remaining 5 XFCC fields.

dhermes:
## PR Summary

 - **Type:** Features

#### Reviewers:

### DIFF
--- a/envoyutil/xfcc.go
+++ b/envoyutil/xfcc.go
@@ -20,8 +20,9 @@ type XFCCElement struct {
 	// `url.URL` if desired.
 	By string
 	// Hash contains the SHA 256 digest of the current client certificate; this
-	// is a bytestring of 64 hexadecimal characters.
-	Hash []byte
+	// is a string of 64 hexadecimal characters. This can be converted to the raw
+	// bytes underlying the hex string via `hex.DecodeString(xe.Hash)`.
+	Hash string
 	// Cert contains the entire client certificate in URL encoded PEM format.
 	// It is present here as a `string` and can be parsed to an `x509.Certificate`
 	// if desired.
@@ -149,7 +150,10 @@ func fillXFCCKeyValue(key, element string, value []rune, ele *XFCCElement) (err 
 		}
 		ele.By = string(value)
 	case "hash":
-		return nil
+		if len(ele.Hash) > 0 {
+			return ex.New(ErrXFCCParsing).WithMessagef("Key already encountered %q", key)
+		}
+		ele.Hash = string(value)
 	case "cert":
 		return nil
 	case "chain":
--- a/envoyutil/xfcc.go
+++ b/envoyutil/xfcc.go
@@ -226,7 +226,7 @@ func fillXFCCKeyValue(key, element string, value []rune, ele *XFCCElement) (err 
 		}
 		ele.URI = string(value)
 	case "dns":
-		return nil
+		ele.DNS = append(ele.DNS, string(value))
 	default:
 		return ex.New(ErrXFCCParsing).WithMessagef("Unknown key %q", key)
 	}
--- a/envoyutil/xfcc.go
+++ b/envoyutil/xfcc.go
@@ -1,10 +1,13 @@
 package envoyutil
 
 import (
+	"crypto/x509"
 	"encoding/hex"
 	"fmt"
+	"net/url"
 	"strings"
 
+	"github.com/blend/go-sdk/certutil"
 	"github.com/blend/go-sdk/ex"
 	"github.com/blend/go-sdk/stringutil"
 )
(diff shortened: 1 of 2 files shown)


envoyutil/xfcc.go
```diff
@@ -39,9 +119,25 @@ func (xe XFCCElement) String() string {
...
+		parts = append(parts, fmt.Sprintf("Hash=%s", maybeQuoted(xe.Hash)))
+	}
+	if xe.Cert != "" {
+		parts = append(parts, fmt.Sprintf("Cert=%s", maybeQuoted(xe.Cert)))
+	}
+	if xe.Chain != "" {
+		parts = append(parts, fmt.Sprintf("Chain=%s", maybeQuoted(xe.Chain)))
+	}
+	if xe.Subject != "" {
+		parts = append(parts, fmt.Sprintf("Subject=%q", xe.Subject))
```
dhermes, line 132: Note here this **always** quoted, from the spec https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#x-forwarded-client-cert

> Subject The Subject field of the current client certificate. The value is always double-quoted.

envoyutil/xfcc.go
```diff
@@ -94,59 +189,69 @@ func ParseXFCCElement(element string) (XFCCElement, error) {
...
-				valueStart = -1
-				valueEnd = -1
+				value = []rune{}
 				state = parseXFCCKey
 			} else {
-				if valueStart == -1 {
-					valueStart = i
-				}
-				valueEnd = i
+				value = append(value, char)
```
dhermes, line 204: This is decently performant and doesn't have the baggage of an accidental runtime panic as a possibility (which the `valueEnd` / `valueStart` approach did have)

See: https://play.golang.org/p/u7eyNdc2Cue

envoyutil/xfcc.go
```diff
@@ -14,12 +18,88 @@ import (
...
+// DecodeChain decodes the `Chain` element from a URL encoded PEM to a
+// `[]x509.Certificate`.
+func (xe XFCCElement) DecodeChain() ([]*x509.Certificate, error) {
+	if xe.Chain == "" {
+		return nil, nil
+	}
+
+	value, err := url.QueryUnescape(xe.Chain)
+	if err != nil {
+		return nil, err
```
mat285, line 94: This error isn't wrapped in an exception right?

envoyutil/xfcc.go
```diff
@@ -14,12 +18,88 @@ import (
...
+	}
+
+	return certutil.ParseCertPEM([]byte(value))
+}
+
+// DecodeURI decodes the `URI` element from a URI string to a `*url.URL`.
+func (xe XFCCElement) DecodeURI() (*url.URL, error) {
+	return url.Parse(xe.URI)
```
mat285, line 102: Should we add some exception wrapping to give us stack traces?

envoyutil/xfcc.go
mat285: We can also make some reasonable pre-allocations when we define `value` at first too if you want, to avoid more mallocs at the cost of eating slightly more memory

mat285: also why not do the same approach the key too instead of the `+=` on the string

dhermes: I actually do that in #212

envoyutil/xfcc.go
dhermes: Good call out. I went through and wrapped all errors returned by the `Decode*()` helpers.

envoyutil/xfcc.go
dhermes: Good call out. I went through and wrapped all errors returned by the `Decode*()` helpers.

envoyutil/xfcc.go
dhermes: > We can also make some reasonable pre-allocations when we define `value` at first too if you want, to avoid more mallocs at the cost of eating slightly more memory

Done. I went with

```go
value := make([]rune, 0,initialValueCapacity)
```

where `initialValueCapacity = 8`.

dhermes: @mat285 I think we should "resolve" this comment even though I didn't do the `key := make([]rune, 0, N)` approach for `key` (since it's coming in #212 and #212 is already done).

mat285: agreed

envoyutil/xfcc.go
```diff
@@ -46,12 +46,22 @@ type XFCCElement struct {
 
 // DecodeBy decodes the `By` element from a URI string to a `*url.URL`.
 func (xe XFCCElement) DecodeBy() (*url.URL, error) {
-	return url.Parse(xe.By)
+	u, err := url.Parse(xe.By)
+	if err != nil {
+		return nil, ex.New(err)
```
mat285, line 51: Nit: Should we use the same error class `ErrorXFCCParsing` and then use `WithInner`?

PR #213, opened 2020-07-30
Adding SPIFFE extraction and middleware helpers for XFCC.

dhermes:
## PR Summary

 - **Type:** Features

#### Reviewers:

### DIFF
--- a/envoyutil/identity.go
+++ b/envoyutil/identity.go
@@ -107,7 +107,7 @@ func SPIFFEClientIdentityProvider(opts ...IdentityProcessorOption) IdentityProvi
 	return processor.IdentityProvider
 }
 
-// ServerIdentityFromSPIFFE produces a verifier function satisfying `VerifyXFCC`.
+// SPIFFEServerIdentityProvider produces a verifier function satisfying `VerifyXFCC`.
 //
 // This function assumes the server identity is in the `By` field and that field
 // is a SPIFFE URI.
@@ -122,7 +122,7 @@ func SPIFFEClientIdentityProvider(opts ...IdentityProcessorOption) IdentityProvi
 //   `IdentityProcessor.KubernetesIdentityFormatter`.
 // - Provide an allow list for the server identity string.
 // - Provide a deny list for the server identity string.
-func ServerIdentityFromSPIFFE(opts ...IdentityProcessorOption) VerifyXFCC {
+func SPIFFEServerIdentityProvider(opts ...IdentityProcessorOption) VerifyXFCC {
 	processor := IdentityProcessor{}
 	for _, opt := range opts {
 		opt(&processor)
--- a/envoyutil/identity.go
+++ b/envoyutil/identity.go
@@ -81,3 +81,28 @@ func ExtractAndVerifyClientIdentity(req *http.Request, cip IdentityProvider, ver
 	// Do final extraction.
 	return cip(xfcc)
 }
+
+// ClientIdentityFromSPIFFE produces a function satisfying `IdentityProvider`.
+//
+// This function assumes the client identity is in the `URI` field and that field
+// is a SPIFFE URI.
+//
+// It delegates processing of that SPIFFE URI via the `IdentityProcessor`
+// type. The options supported can
+// - Provide an allow list for the trust domain in the SPIFFE URI.
+// - Provide a deny list for the trust domain in the SPIFFE URI.
+// - Provide a function to produce a client identity string from the SPIFFE
+//   URI (likely from the workload ID in the SPIFFE URI); if no option is
+//   provided for this the default will use
+//   `IdentityProcessor.KubernetesIdentityFormatter`.
+// - Provide an allow list for the client identity string.
+// - Provide a deny list for the client identity string.
+func ClientIdentityFromSPIFFE(opts ...IdentityProcessorOption) IdentityProvider {
+	processor := IdentityProcessor{}
+	for _, opt := range opts {
+		opt(&processor)
+	}
+	// Ensure the `Type` is "client" even if `opts` set it to be otherwise.
+	processor.Type = ClientIdentity
+	return processor.IdentityProvider
+}
--- a/envoyutil/identity.go
+++ b/envoyutil/identity.go
@@ -41,9 +41,10 @@ type IdentityProvider func(xfcc XFCCElement) (identity string, err error)
 type VerifyXFCC func(xfcc XFCCElement) error
 
 // ExtractAndVerifyClientIdentity enables extracting client identity from a request.
-// It does so by requiring the XFCC header to present and valid and contain exactly
-// one element. Then it passes the parsed XFCC header along to some verifiers (e.g.
-// to verify the server identity) as well as to an extractor (for the client identity).
+// It does so by requiring the XFCC header to be present and valid and contain exactly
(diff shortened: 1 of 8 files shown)


envoyutil/identity.go
```diff
@@ -7,40 +7,53 @@ import (
 )
 
 const (
-	// ErrMissingXFCC is the error returned when XFCC is missing
+	// ErrMissingXFCC is the error returned when XFCC is missing.
```
dhermes, line 10: Reviewer(s): **note** this PR is based on #208 

I started adding the middleware in the `dhermes/client-identity-add-middleware` branch and realized a lot of pieces in #208 needed to be slightly re-worked.

In this PR the commit history is completely re-done to "make sense" and slowly build up the functions and types as behavior is added.

envoyutil/identity.go
```diff
@@ -68,3 +81,57 @@ func ExtractAndVerifyClientIdentity(req *http.Request, cip ClientIdentityProvide
 	// Do final extraction.
 	return cip(xfcc)
 }
+
+// ClientIdentityFromSPIFFE produces a function satisfying `IdentityProvider`.
```
wcharczuk, line 85: SpiffeIdentityProvider?

envoyutil/identity.go
```diff
@@ -68,3 +81,57 @@ func ExtractAndVerifyClientIdentity(req *http.Request, cip ClientIdentityProvide
...
+// type. The options supported can
+// - Provide an allow list for the trust domain in the SPIFFE URI.
+// - Provide a deny list for the trust domain in the SPIFFE URI.
+// - Provide a function to produce a server identity string from the SPIFFE
+//   URI (likely from the workload ID in the SPIFFE URI); if no option is
+//   provided for this the default will use
+//   `IdentityProcessor.KubernetesIdentityFormatter`.
+// - Provide an allow list for the server identity string.
+// - Provide a deny list for the server identity string.
+func ServerIdentityFromSPIFFE(opts ...IdentityProcessorOption) VerifyXFCC {
```
wcharczuk, line 125: same nit with the client identity provider here

envoyutil/identity.go
```diff
@@ -68,3 +81,57 @@ func ExtractAndVerifyClientIdentity(req *http.Request, cip ClientIdentityProvide
...
+	processor := IdentityProcessor{}
+	for _, opt := range opts {
+		opt(&processor)
+	}
+	// Ensure the `Type` is "server" even if `opts` set it to be otherwise.
+	processor.Type = ServerIdentity
```
wcharczuk, line 131: what i generally do here is initialize the thing i'm mutating with some options vs. getting this adversarial but i understand why it's critical to set

wcharczuk: no specific action here just insight

envoyutil/identity_processor.go
```diff
@@ -0,0 +1,288 @@
...
+	"github.com/blend/go-sdk/collections"
+	"github.com/blend/go-sdk/ex"
+	"github.com/blend/go-sdk/spiffeutil"
+)
+
+// NOTE: Ensure that
+//       - `IdentityProcessor.KubernetesIdentityFormatter` satisfies `IdentityFormatter`
+//       - `IdentityProcessor.IdentityProvider` satisfies `IdentityProvider`
+var (
+	_ IdentityFormatter = IdentityProcessor{}.KubernetesIdentityFormatter
```
wcharczuk, line 16: not a fan of allocating 2x structs to do these sniffs, can we use `((*IdentityProcessor)(nil)).KubernetesIdentityFormatter`?

envoyutil/identity_processor.go
```diff
@@ -0,0 +1,288 @@
...
+
+// IdentityProcessor provides configurable fields that can be used to
+// help validate a parsed SPIFFE URI and produce and validate an identity from
+// a parsed SPIFFE URI. The `Type` field determines if a client or server
+// identity should be provided; by default the type will be client identity.
+type IdentityProcessor struct {
+	Type                IdentityType
+	AllowedTrustDomains []string
+	DeniedTrustDomains  []string
+	AllowedIdentities   collections.SetOfString
```
wcharczuk, line 92: GENERICS WHEN

envoyutil/identity.go
dhermes: Done

envoyutil/identity.go
dhermes: Done

envoyutil/identity.go
dhermes: Yeah I'm not that happy about having to do it, but I did it in the name of not surprising users if a thing that claims to get a server identity gets a client identity (even if it was user error to configure it differently).

I could introduce something like

```go
// NOTE: Ensure that
//       - `expectedServerIdentity` satisfies `VerifyXFCC`
var (
	_ VerifyXFCC = expectedServerIdentity
)

func expectedServerIdentity(_ XFCCElement) error {
	return ex.New("Expected server identity")
}
```

and then replace this line and comment with

```go
	// If the `Type` is not "server", we can't use `processor`.
	if processor.Type != ServerIdentity {
		return expectedServerIdentity
	}
```

dhermes: We could also just go the route of `processor := IdentityProcessor{Type: ServerIdentity}` and making the `godoc` extra explicit about users overriding the `Type`.

envoyutil/identity_processor.go
dhermes: I agree. I initially did `(*IdentityProcessor)(nil).KubernetesIdentityFormatter` but it panics because the receiver in `KubernetesIdentityFormatter` is a value receiver, e.g.

```diff
diff --git a/envoyutil/identity_processor.go b/envoyutil/identity_processor.go
index e1fb936..8e7da6d 100644
--- a/envoyutil/identity_processor.go
+++ b/envoyutil/identity_processor.go
@@ -13,7 +13,7 @@ import (
 //       - `IdentityProcessor.KubernetesIdentityFormatter` satisfies `IdentityFormatter`
 //       - `IdentityProcessor.IdentityProvider` satisfies `IdentityProvider`
 var (
-	_ IdentityFormatter = IdentityProcessor{}.KubernetesIdentityFormatter
+	_ IdentityFormatter = (*IdentityProcessor)(nil).KubernetesIdentityFormatter
 	_ IdentityProvider  = IdentityProcessor{}.IdentityProvider
 )
 
```

yields

```
$ go test
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1447962]

goroutine 1 [running]:
github.com/blend/go-sdk/envoyutil.init()
        /Users/dhermes/go/src/github.com/blend/go-sdk/envoyutil/identity_processor.go:16 +0x42
exit status 2
FAIL    github.com/blend/go-sdk/envoyutil       0.247s
```

----

We have at least three options here:

- Leave it be / eat the cost (not sure if the compiler tries to save us from the allocation cost)
- I can change `KubernetesIdentityFormatter` to take a pointer receiver
- I can move the cost of creating the `IdentityProcessor{}` to a unit test (at runtime of the unit test)

wcharczuk: the way it is is fine for now

PR #218, opened 2020-08-25
Updating `configutil` and all usage of it.

dhermes:
## PR Summary

 - **Type:** Internal

#### Reviewers:


configutil/int64_test.go
```diff
@@ -0,0 +1,23 @@
+package configutil
+
+import (
+	"context"
+	"testing"
+
+	"github.com/blend/go-sdk/assert"
+)
+
+func TestInt646464646464(t *testing.T) {
```
wcharczuk, line 10: hey is this the name from the monorepo?

wcharczuk: i have confirmed, it is in fact, the name from the monorepo

wcharczuk, line 10:
```suggestion
func TestInt64(t *testing.T) {
```

configutil/env.go
```diff
@@ -133,3 +100,11 @@ func (e EnvVars) Duration() (*time.Duration, error) {
 	}
 	return nil, nil
 }
+
+// vars resolves the env vars
+func (e EnvVars) vars(ctx context.Context) env.Vars {
+	if vars := env.GetVars(ctx); vars != nil {
```
wcharczuk, line 106: did this function make it with the rest of the changes? I can't find it

dhermes: I presume "this function" == `env.GetVars()` here? If yes, I pulled in `env` changes in https://github.com/blend/go-sdk/pull/215.

configutil/file.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+// File reads the string contents of a file as a literal config value.
+type File string
+
+// String returns the string contents of a file.
+func (f File) String(ctx context.Context) (*string, error) {
+	contents, err := ioutil.ReadFile(string(f))
+	if err != nil {
+		return nil, nil
```
mat285, line 16: Not returning the error?

dhermes: This is a copy-pasted port of existing code in our internal version of this. Happy to patch bugs in the implementation but would strongly prefer to do in a different PR.

wcharczuk: yeah it should ideally differentiate between `ENOTFOUND` and an actual error reading from disk, but for now we just skip over the variable source if there is any issue vs. failing resolving the config

PR #228, opened 2020-10-23
Remove cookies from event sent to sentry

chriskim06:
## PR Summary

 - **Type:** Improvements

#### Reviewers:

### DIFF
--- a/sentry/client.go
+++ b/sentry/client.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/blend/go-sdk/ex"
 	"github.com/blend/go-sdk/logger"
+	"github.com/blend/go-sdk/webutil"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


sentry/client.go
```diff
@@ -109,6 +109,9 @@ func errRequest(ee logger.ErrorEvent) *raven.Request {
 		return &raven.Request{}
 	}
 
+	// remove cookies from the error event if they exist
+	typed.Header.Del("Cookie")
```
dhermes, line 113: We may want to borrow a constant from elsewhere in `go-sdk` (rather than `"Cookie"`)

- https://github.com/blend/go-sdk/blob/1bd31746db880e1ff4be3689fd354adfacbae8ec/webutil/constants.go#L28
- https://github.com/blend/go-sdk/blob/1bd31746db880e1ff4be3689fd354adfacbae8ec/web/constants.go#L30

(it's less likely to be "fun" to import `go-sdk/web`, it's a heavier dependency)

----

Is it OK to modify `typed.Header`? Since it's a pointer, the caller may still have a reference to this so we shouldn't modify it?

sentry/client_test.go
```diff
@@ -59,6 +65,7 @@ func TestErrEvent(t *testing.T) {
 	assert.NotEmpty(event.Fingerprint)
 	assert.NotNil(event.Exception[0].Stacktrace)
 	assert.NotEmpty(event.Exception[0].Stacktrace.Frames)
+	assert.Equal("", event.Request.Cookies)
```
dhermes, line 68: Can you also do an assertion on `req.Header.Get("Cookie")`?

sentry/client.go
chriskim06: Ya you're right I didn't really notice that. I could modify the sentry request that gets created instead of the http request.

dhermes (approved):
LGTM

PR #231, opened 2020-11-11
Adding `db.Config.Application` field to set the `application_name` configuration parameter

dhermes:
## PR Summary

 - **Type:** Features

#### Reviewers:

### DIFF
--- a/db/config.go
+++ b/db/config.go
@@ -63,23 +63,23 @@ func NewConfigFromDSN(dsn string) (*Config, error) {
 
 // NewConfigFromEnv returns a new config from the environment.
 // The environment variable mappings are as follows:
-// 	-	DB_ENGINE            = Engine
-//	-	DATABASE_URL         = DSN     //note that this has precedence over other vars (!!)
-// 	-	DB_HOST              = Host
-//	-	DB_PORT              = Port
-//	- 	DB_NAME              = Database
-//	-	DB_SCHEMA            = Schema
-//	-	DB_APPLICATION_NAME  = ApplicationName
-//	-	DB_USER              = Username
-//	-	DB_PASSWORD          = Password
-//	-	DB_CONNECT_TIMEOUT   = ConnectTimeout
-//	-	DB_LOCK_TIMEOUT      = LockTimeout
-//	-	DB_STATEMENT_TIMEOUT = StatementTimeout
-//	-	DB_SSLMODE           = SSLMode
-//	-	DB_IDLE_CONNECTIONS  = IdleConnections
-//	-	DB_MAX_CONNECTIONS   = MaxConnections
-//	-	DB_MAX_LIFETIME      = MaxLifetime
-//	-	DB_BUFFER_POOL_SIZE  = BufferPoolSize
+//  -  DB_ENGINE            = Engine
+//  -  DATABASE_URL         = DSN     //note that this has precedence over other vars (!!)
+//  -  DB_HOST              = Host
+//  -  DB_PORT              = Port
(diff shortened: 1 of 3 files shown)


db/config.go
```diff
@@ -38,6 +38,8 @@ func NewConfigFromDSN(dsn string) (*Config, error) {
 			config.SSLMode = strings.TrimPrefix(piece, "sslmode=")
 		} else if strings.HasPrefix(piece, "search_path=") {
 			config.Schema = strings.TrimPrefix(piece, "search_path=")
+		} else if strings.HasPrefix(piece, "application_name=") {
```
dhermes, line 41: See:

- https://www.postgresql.org/docs/12/runtime-config-logging.html#GUC-APPLICATION-NAME
- https://www.datadoghq.com/blog/collect-postgresql-data-with-datadog/#collecting-custom-postgresql-metrics-with-datadog

dhermes: I plan on using a Kubernetes pod name as the `application_name` to uniquely identify connection count (and Idle / Active state) within Datadog monitors

db/config.go
```diff
@@ -61,22 +63,23 @@ func NewConfigFromDSN(dsn string) (*Config, error) {
...
-//	-	DB_MAX_CONNECTIONS   = MaxConnections
-//	-	DB_MAX_LIFETIME      = MaxLifetime
-//	-	DB_BUFFER_POOL_SIZE  = BufferPoolSize
+//  -  DB_ENGINE            = Engine
+//  -  DATABASE_URL         = DSN     //note that this has precedence over other vars (!!)
+//  -  DB_HOST              = Host
+//  -  DB_PORT              = Port
+//  -  DB_NAME              = Database
+//  -  DB_SCHEMA            = Schema
+//  -  DB_APPLICATION_NAME  = ApplicationName
```
wcharczuk, line 72: `Application`? the `Name` feels redundant, what do you think?

wcharczuk: Strictly speaking, that is the name in the underlying driver dsn so i'm fine with it.

wcharczuk (approved):
LGTM, plz port to golang.blend.com

db/config.go
wcharczuk: yeah makes sense

db/config.go
dhermes: I'm OK with that. I chose it because the `libpq` parameter is `application_name`

db/config.go
wcharczuk: is this field supported in postgres < 12?

dhermes:
> LGTM, plz port to golang.blend.com

Totally! (Definitely was planning on it)

db/config.go
wcharczuk: yeah lets keep it at `application_name` for now

db/config.go
dhermes: - https://www.postgresql.org/docs/9.5/runtime-config-logging.html#GUC-APPLICATION-NAME
- https://www.postgresql.org/docs/9.6/runtime-config-logging.html#GUC-APPLICATION-NAME
- https://www.postgresql.org/docs/10/runtime-config-logging.html#GUC-APPLICATION-NAME
- https://www.postgresql.org/docs/11/runtime-config-logging.html#GUC-APPLICATION-NAME

AFAICT it was introduced in 9.0, but earlier than 9.5 is completely unsupported so we're all good with 9.5+

db/config.go
dhermes: SGTM

PR #234, opened 2020-12-04
Initial golang.blend.com export

wcharczuk:


assert/.profanity.yml
```diff
@@ -0,0 +1,4 @@
+GOLANG_ASSERT_GO_SDK:
+  description: "sdk/assert needs to only depend on itself as it's used by every other package"
+  goImports: 
+    include: [ "golang.blend.com/sdk/*" ]
```
wcharczuk, line 4: we should probably exclude profanity files

dhermes (approved):

PR #242, opened 2021-07-13
use BLAKE3 as Hash algo choice

ykcab:
## PR Summary

 - **Type:** Improvements

#### Reviewers:

- Does the "Intended Change Level" above match the actual behavior of this PR? Yes
- Is this PR following patterns set forth in the package (if modifying a package), or the go-sdk design patterns if implementing a new package from scratch?  N/A
- Does this require a backport to LTS versions? N/A

### DIFF
--- a/crypto/local_transit.go
+++ b/crypto/local_transit.go
@@ -82,12 +82,16 @@ func (m LocalTransit) Encrypt(dst io.Writer, src io.Reader) error {
 	if m.ContextProvider != nil {
 		currentContext = m.ContextProvider()
 	}
-	key, err := m.GetKey(currentContext)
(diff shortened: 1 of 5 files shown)


ykcab:
I noticed that both the HMAC and Encryption use the same key. It is not a secure design. This is the same behavior noticed in the previous design.

Note: In this PR, the local transit logic needs some review to possibly pass both the HMAC key and Encryption key.

wcharczuk:
Hey @ykcab thanks for the PR! we don't accept PRs on this repo currently, I've made some edits to the README.md to that effect, and incorporated the hmac split key tweak for the public api (note: we don't use this internally, we use something else, but will be helpful if others do). Going to close this out for now.

